PR #1, opened 2025-05-26
Design Proposal: Initial Concept for Expression Purity Checker

HenrikPaetzold:


Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
+## !First Concept! for the Purity Checker
```
jesyspa: There's no need to emphasise this, documents evolve over time

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+We want a call of `checkValidity()` on an arbitrary `ExpEmbedding` node to return `true` if all `assert` statements present contain only pure expressions, and `false` as soon as one of them is impure. We split the logic into traversal and validation, currently only focussing on the validation of expressions appearing inside of `assert` nodes.
```
jesyspa: This is very specific.  What we want is to introduce a notion of validity that is established after the translation to ExpEmbedding, and assert statements with impure conditions are an example of a violation.

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+Every `ExpEmbedding` implements a new `children` property and the functions `checkValidity` and `checkOwnValidity`.
```
jesyspa: I wouldn't introduce something new; rather, we have a `subexpressions` property on many ExpEmbeddings already.

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+The default for `children` is an empty immutable list (`emptyList()`); when overridden it must still be an immutable list referencing the children’s subsequent embeddings.
```
jesyspa: This is very specific, and I wouldn't be surprised if it wasn't what we eventually wanted; e.g. a sequence of children may make more sense.

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Only validating `Assert`
+
+```kotlin
+data class Assert(val exp: ExpEmbedding) : ExpEmbedding {
+
+    override fun children(): List<ExpEmbedding> = listOf(exp)
+
+    override fun checkValidity(): Boolean =
+        checkOwnValidity()                       // stop traversal here
```
jesyspa: Why?  Wouldn't we want to report validity errors from subexpressions?

Design/purity-checker-proposal.md
jesyspa: Apart from returning true/false, I guess we also want to report the validity issue to the user?

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+
+An `Assert` node ends the traversal and validates only itself.
+
+---
+
+### PurityRules
+
+`assert` can only contain embeddings evaluating to a boolean, so for the validation we can restrict ourselves to:
```
jesyspa: Does it matter?  The subexpressions of these can still evaluate to something else.

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+            val calleePure = e.method.name in pureSpecialFunctions
+            calleePure && e.args.all(::isExpressionValid)
+        }
+        
+        // always forbidden inside verify(...)
+        is Assert,
+        is Assign,
+        is FieldModification        -> false
+
+        else -> e.children().all(::isExpressionValid)   // covers DirectResult, etc.
```
jesyspa: This seems rather broad, why does it make sense?

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+```
+
+By splitting traversal and validation like this, adapting to future changes (e.g., pure-method annotations) should be simpler than with the previous purity-checker class. I opted to implement `children` for all nodes so that individual `checkValidity` overrides are rarely needed, and because the whole mechanism is stateless we can simply call `checkValidity` on the root of the tree.
+
+### Side Notes
+
+I am assuming that I don’t need to check whether asserts only contain BooleanTypeEmbeddings, because it already throws errors, when something doesn’t evaluate to a boolean and we check purity, not correctness
```
jesyspa: I don't think checking type correctness needs to be part of this proposal, though it would make sense as a sanity check at some point...

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+It might not work for strings right now. I also haven’t found out how to check which operation is used inside an assert (≤,<, ==) but I honestly think, that that wouldn’t be necessary since we don’t check the type anyway.
```
jesyspa: I think these are all binary operator expembeddings?  I doubt you need to know the specific one, though.

jesyspa (changes requested):
Generally looks reasonable, some comments

Design/purity-checker-proposal.md
HenrikPaetzold: Yes, I noticed that. I considered two possible approaches for handling the child nodes: either each class implements a children property, allowing the default implementation of checkValidity to work, or each class overrides checkValidity directly with its specific child logic. I opted for the first approach, but am also fine with the second one.

Design/purity-checker-proposal.md
HenrikPaetzold: This didn’t occur to me yesterday, but after the meeting, I realized it might be beneficial. For now, as discussed, I will leave it as is.

Design/purity-checker-proposal.md
HenrikPaetzold: Yes, absolutely

Design/purity-checker-proposal.md
HenrikPaetzold: No, it doesn’t. And I don’t think that sentence should have been there at all, since it doesn’t align with the behavior described below. We should be traversing every node in the tree, and subexpressions can of course be of types other than boolean.

Design/purity-checker-proposal.md
HenrikPaetzold: I’m not really sure what I should say here, so maybe it’s best to explain what I was thinking when I wrote this:

The idea behind this method—and specifically the added recursion—was based on the assumption that we only call it once everything has already been converted into a correct exp-embedding representation that makes sense in the current state of the plugin.

We invoke this method when we encounter an assert embedding, and at that point, we assume the subtree it contains has already been correctly converted—if something were fundamentally wrong with the expression, the program would have already thrown an error. Based on that assumption, checking for purity becomes a matter of traversing the rest of the tree: either we reach a leaf node and determine whether it's pure or impure, or we reach an internal node with children. In the latter case, we need to verify the purity of all those children (which was the reason I initially introduced the children property). I did consider the idea that a subtree might be impure while its parent remains pure, but that doesn’t really hold—consider a logical AND, which is exactly what the else branch is meant to handle.

With the children property gone, I think the best approach is to explicitly handle each of the different embedding types that could appear. Relying on the else branch to cover a large subset of nodes might not be good practice. I hope this made sense.
In case I understood you wrong—could you maybe specify what you don’t like about this?
Otherwise, I’ll get started on an implementation.

Design/purity-checker-proposal.md
jesyspa: Staring at this for a bit more, I think a `children` node is fine, though I'd recommend a different way to traverse the whole thing: instead of having `checkValidity` do this work, let's make a `preorderTraversal` function that returns a sequence, so we can easily write any algorithm that hits every node?

Design/purity-checker-proposal.md
jesyspa: > Based on that assumption, checking for purity becomes a matter of traversing the rest of the tree: either we reach a leaf node and determine whether it's pure or impure, or we reach an internal node with children.

This is the key point here.  I am not sure this assumption is valid: is every node that isn't one of the above pure precisely when its children are pure?

Design/purity-checker-proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+### PurityRules
+
+`assert` can only contain embeddings evaluating to a boolean, so for the validation we can restrict ourselves to:
+
+```kotlin
+private val pureSpecialFunctions: Set<MangledName>; //place the pure Subset of the FullySpecialKotlinFunctions here
+
+object PurityRules {
+    fun isExpressionValid(e: ExpEmbedding): Boolean = when (e) {
```
jesyspa: This has a similar issue as the validity check earlier.  If you'd like to traverse a heterogeneous tree, I would strongly recommend either writing a visitor or making it a virtual function.

jesyspa: I think this should be called `isExpressionPure`, by the way?

jesyspa (approved):
LGTM, though I would recommend trying to phrase things in text rather than just putting the code: the goal of a design doc is that it should be easier to understand than looking at the code would be.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+
+### checkValidity
+
+To check the validity of the graph, we obtain it using the `preorder` function. We filter all embeddings for which we want to check validity and call `isValid()` on them.
+
+```kotlin
+fun ExpEmbedding.checkValidity(): Boolean =
+    preorder()
+        .filterIsInstance<Assert>()
+        .all { it.isValid() }
```
jesyspa: Why not call `isValid` on everything, and then just have the default implementation be a no-op?  This will make it easier to e.g. add support for pre-/postconditions.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+    } 
+        
+}
+``` 
+
+---
+
+### Functionality of `ExprPurityVisitor`
+
+The `ExprPurityVisitor` is responsible for traversing `Assert` expressions and verifying their purity. It should reject any construct that either cannot appear inside an `assert` or is impure (currently `return false`). At the same time, it should accept expressions that are pure in the base cases (`return true`) and recursively traverse the expression tree down to its leaves where necessary.
```
jesyspa: The purity checker shouldn't care about where it was called: it should just tell you whether a certain piece of code is pure or not.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+
+Sure! Here's the polished English version in a code block for easy copying:
+
+Got it — you liked the visual style of the indentation, just not the fact that Markdown ignores it.
+
+To preserve the indentation *and* make it render properly, we can wrap it in a **code block**. That way, the indentation is preserved exactly as you intended, and it's still readable in Markdown renderers like GitHub or GitLab.
+
+Here’s the version with indentation that **renders as-is**:
```
jesyspa: I once again urge you to write your own documents.  Thinking your design through as you write it down is very helpful to spotting the issues.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+[Purity] verify(true) - expression is pure
+```
+
+Since the `DiagnosticReporter` is not available in `isValid`, we decided it's unnecessary to propagate all the data required for a `reportOn()` call—especially when the only varying components are the `expr` and whether or not it is pure. Instead, we define a lambda:
+
+```
+reportPurityDiag: (String, String) -> Unit
+```
+
+This allows us to pass only the essential information, while the surrounding context (which doesn't change) remains fixed within the `ViperPoweredDeclarationChecker`.
```
jesyspa: I don't know what you mean with this sentence.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+
+```
+[Purity] verify(true) - expression is pure
+```
+
+Since the `DiagnosticReporter` is not available in `isValid`, we decided it's unnecessary to propagate all the data required for a `reportOn()` call—especially when the only varying components are the `expr` and whether or not it is pure. Instead, we define a lambda:
+
+```
+reportPurityDiag: (String, String) -> Unit
+```
```
jesyspa: As discussed, please put such a function in the ErrorReporter and propagate that.

Design/purity-checker.md
```diff
@@ -0,0 +1,107 @@
...
+
+This allows us to pass only the essential information, while the surrounding context (which doesn't change) remains fixed within the `ViperPoweredDeclarationChecker`.
+
+To propagate this lambda, we’ve opted to use `context` parameters—function signatures are already cluttered in places, and no single function other than `Assert.isValid()` actually needs `reportPurityDiag`.
+
+To access a clear-text version of the expression inside an `Assert`, we take advantage of the fact that `FirStatement` elements are wrapped in a `withPosition` container during conversion. This wrapper stores useful source information alongside the corresponding `ExpEmbedding`. Using it, we can retrieve the original source representation via `getElementTextInContextForDebug()`, which allows us to generate a meaningful error message.
```
jesyspa: One thing to note is that you probably do not want to actually access the text of the violation.  "Assert condition is impure" is fine: the error will already be attached to the right place in the source.

jesyspa (changes requested):

Design/purity-checker.md
HenrikPaetzold: Yes, absolutely. I had some issues with markdown.

Design/purity-checker.md
```diff
@@ -0,0 +1,87 @@
...
+Every `ExpEmbedding` implements a new `children` property (`:Sequence<ExpEmbedding>`) and the function `isValid()`. 
+
+---
+
+### Preorder
+
+We define an extension function `ExpEmbedding.preorder()` that traverses the `ExpEmbedding` tree and returns an iterable sequence in the graph's preorder. SnaKt stores information about FIR source elements by encapsulating an embedding in a `WithPosition` node. To enable accurate positioning in diagnostic reporting, we introduce a new helper:
+```kotlin
+class PositionedExpEmbedding(val embedding: ExpEmbedding, val source: KtSourceElement?)
+```
```
jesyspa, line 16: 50:50 on this since it's just a `WithPosition` node, but I think it's fine.

jesyspa (approved):

PR #2, opened 2025-08-27
Create new-mangling-approach.md

braulov:


Design/new-mangling-approach.md
jesyspa, line 1: nit: `new` is a dangerous word to use in a name, since things quickly become old.  `name-mangling` is fine.  (If there's already such a file, move it to an `outdated` directory and mark it with the date it became outdated.)

Design/new-mangling-approach.md
```diff
@@ -0,0 +1,74 @@
+## Name Mangling
+
```
jesyspa: Start by stating the goals and constraints.

Design/new-mangling-approach.md
```diff
@@ -0,0 +1,74 @@
...
+### 1. Registration of Types
+
+- Types are registered to track unique identifiers for variables.
```
jesyspa: If you just say `type`, the assumption is that it's about Kotlin types.  Explain the structure of names.

Design/new-mangling-approach.md
```diff
@@ -0,0 +1,74 @@
...
+
+### 2. Registration of Names
+
+- Names (e.g., variable basenames) are registered within scopes.
+- Each name is associated with a type, allowing for type-based disambiguation during resolution.
+
+### 3. Assigning Output Versions to Types
+
+- For each type, select a single output version from $\{short, medium, long\}$ used for all variables of that type.
```
jesyspa: Why is this desireable?  It seems like the key slicing should be by name: for a name `x` we may decide on different strategies for disambiguating it (by type, by scope, by argument (Kotlin) types).  Why make this choice per type?

jesyspa (changes requested):
Looks like a good start, but I'd like to address some points before reviewing the rest, feel free to book a meeting if you'd rather discuss in person

Design/name-mangling.md
```diff
@@ -1,188 +1,20 @@
-# Name Mangling
+## Name mangling
 
-When converting FIR to Viper, we need to ensure that names
-do not conflict.  However, we would prefer short names over
-long names.  By registering all names used in the program
-and attempting to find cases when they can be shortened
-without collisions, we can make the generated code much more
-readable.
+When converting FIR to Viper, we need to ensure that names do not conflict. However, we would prefer short names over long names. By registering all names used in the program and attempting to find cases when they can be shortened without collisions, we can make the generated code much more readable.
```
jesyspa: Short line lengths were intentional, it's the same when rendered but easier to read when it isn't.

Design/name-mangling.md
```diff
@@ -1,188 +1,20 @@
-# Name Mangling
```
jesyspa, line 1: Why?

Design/name-mangling.md
```diff
@@ -1,188 +1,20 @@
...
-itself has a namespace.
-
-Local names may also contain names: for example, the name of
-a function may contain the types of its parameters.
-However, to make the program tractable we assume that local
-names are guaranteed to be unique in their full form.  Local
-names may also have shorter forms, that should be used if
-they do not collide.
-
-### Example
```
jesyspa, line 34: An example would be beneficial, perhaps you can adapt this one instead of removing it?

Design/name-mangling.md
```diff
@@ -1,188 +1,20 @@
...
-```
+Algorithm:
 
-We could run the algorithm again on namespace-name pairs,
-and this may be worth doing to remove bulky namespace names
-when they aren't necessary.  However, more thought is needed
-to see when that will make life easier.
+1. Traverse the program and add mangled names.
+2. Build a directed dependency graph where each vertex is a pair $(x, \text{version})$, with $x$ being a mangled name and $\text{version} \in \{\text{short}, \text{medium}, \text{long}\}$. Add an edge $(x, v_x) \to (y, v_y)$ if representing $(x, v_x)$ requires knowing the representation of $(y, v_y)$. To check this, we simply analyze the specific cases.
+3. Obtain a topological sort of the dependency graph (given the nature of dependencies on mangled name components, this graph should be acyclic).
```
jesyspa: This suggests names are resolved one-by-one, but how do you ensure that if you have `fun$x` and `local$x` that you don't assign either the name `x`?

braulov: I have this line in the doc: “If there are no conflicts in the graph.”, but I mean “If there are no conflicts with already assigned names.”. Sorry, I’ll update this in the next commit

jesyspa: I don't really get what that means—I thought the point of the toposort was so that you could detect conflicts incrementally, but if you're checking for them globally, what role does it play?

braulov: Topological sort ensures we process names in dependency order, so we know the necessary parts of other names (like a scope, basename) required to define our name.

Design/name-mangling.md
```diff
@@ -1,35 +1,52 @@
...
+initially a string but a symbolic value.
+3. *Type* — a string that indicates the category or kind of the name.
+
+Note that *basename* and *scope* may themselves contain **mangled names**: 
+for example, a function may be namespaced within a class, which
+itself has a scope.
+
+From each **mangled name**, a corresponding **Viper name** in one 
+of three possible forms:
+1. ```basename_reqiredScope```
```
jesyspa: Shouldn't scope come before the basename?  We want `C_f`, not `f_C`, for `class C { fun f() { ... } }`

Design/name-mangling.md
```diff
@@ -1,35 +1,52 @@
...
+2. ```type_basename_requiredScope```
+3. ```type_basename_fullScope```
+
+Where:  
+**fullScope** — string representation of *Scope*.  
+**requiredScope** — string representation of the essential part 
+of *Scope* (currently, only classes).  
+**optionalScope** — string representation of all remaining parts 
+of *Scope*, including the *Type* and any non-essential elements.
```
jesyspa: You don't use `optionalScope` anywhere.

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
+We work with two graphs:
 
-// A: A, fun_A, fun_A_takes_Unit, fun_A_takes_Unit_returns_Unit
-fun A() { ... }
+a **dependency graph** whose vertices are pairs 
+$(x, \text{version})$,  where $x$ is a mangled name 
+and $\text{version} \in \{\text{short}, \text{medium}, 
+\text{long}, \ldots\}$.  
+Edges encode that rendering $(x, v_x)$ requires the representation 
+of $(y, v_y)$.
```
jesyspa: But does `(x, v_x)` know which representation it needs?  Like, doesn't it just depend on `y`?

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
+   The assignment is allowed iff **all** of the following hold:
+  - **No collisions** are created with already assigned strings.
+  - **Basename fairness:** no other name with the **same basename** has
+    already been assigned a **strictly shorter** version than $v$.
+  - **Feasibility:** after the assignment, there is **no** vertex in
+    the conflict graph for which **all** versions become banned
+    (i.e., every vertex still has at least one admissible version).
 
-fun fun_A() { ... }
+   **Group assignment by basename.**
```
jesyspa: I feel like this isn't flexible enough.
Consider the following:

```kotlin
package P {
  class C { fun f() = ... }
}
package R {
  class C { fun f() = ... }
}

package S {
  fun f() = ...
}
```

Now it may happen that `S.f` is picked first.  You check that it does not need disambiguating: it can just be `f`.  But now you've assigned `P.C.f` and `R.C.f` their shortest name, which is `C.f` for both, and you have a collision.

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
-class B {
-    fun B_prop_x_getter(): Int { ... }
-    fun B_prop_x_setter(param_v: Int) { ... }
-}
+2. **Fix the version indicated by the sink vertex.**  
+   For each sink $(x, v)$, we attempt to **assign version $v$** to $x$.  
+   The assignment is allowed iff **all** of the following hold:
+  - **No collisions** are created with already assigned strings.
+  - **Basename fairness:** no other name with the **same basename** has
+    already been assigned a **strictly shorter** version than $v$.
```
jesyspa: Doesn't this contradicts your group assignment requirement? If you assign things by base names, no other name with the same basename can have an assignment yet.

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
+  - **Feasibility:** after the assignment, there is **no** vertex in
+    the conflict graph for which **all** versions become banned
+    (i.e., every vertex still has at least one admissible version).
```
jesyspa: You need to elaborate on what the update of the conflict graph looks like.

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
+   After committing, update the affected subtrees in the dependency graph.
+
+3. **Prune resolved vertices.**  
+   Remove **all vertices whose basename got fixed** (i.e., all
+   $(y, \cdot)$ with $y.\!basename = x.\!basename$), together
+   with incident edges.
+
+5. **Repeat until stable.**  
+   Return to Step 1 while conflicts remain or further assignments
+   are possible.
```
jesyspa: Could you elaborate on what each of these cases means?

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
+Remove all `"foo()"` vertices.
+
+
+#### Iteration 4 — possible fallback later
+
+If later a collision is detected in the conflict graph (e.g. two 
+unrelated `foo()` collapsing to the same string),  
+the algorithm will eventually reach $(foo(), medium)$ as a sink, 
+and then reassign the entire basename-group `"foo()"` to $medium$:
+- `"fun_foo()_A"`, `"fun_foo()"`, …
```
jesyspa: I don't see this in the description of the algorithm?

Design/name-mangling.md
```diff
@@ -78,111 +89,149 @@ We can identify the possible local names in this example:
...
-fun fun_foo_takes_Unit() {
-    val param_x: Int = ...
-}
-```
 
-We could run the algorithm again on namespace-name pairs,
-and this may be worth doing to remove bulky namespace names
-when they aren't necessary.  However, more thought is needed
-to see when that will make life easier.
+Thus, the algorithm never “tries shorter first by hand”:
```
jesyspa: I don't get what you mean by this

Design/name-mangling.md
braulov: Yes, of course, it's just typo.

Design/name-mangling.md
```diff
@@ -78,111 +71,55 @@ We can identify the possible local names in this example:
...
+
+We maintain a **dependence graph** G over names. There is a 
+directed edge $x \to y$ iff determining the string representation 
+of $x$ requires the string representation of $y$.  
+$G$ is a DAG (see **Constraints**).
+
+### Initialization 
+1. For each name ```x```, build its candidate list.
+2. Build ```G```. 
+3. Compute initial strings for all names in a topological order of ```G```.
```
jesyspa: What do you mean by `initial strings`?

Design/name-mangling.md
```diff
@@ -78,111 +71,55 @@ We can identify the possible local names in this example:
...
+```FixSet = chooseNamesForFix(Conflicted)```
+where $\text{FixSet} \subseteq \text{Conflicted}, \text{FixSet}
+\neq \emptyset$
+3. **Advance candidates**  
+For each $x \in \text{FixSet}$, call ```deleteCurrentCandidate(x)```
+to move to the next candidate (or fallback).
+4. **Recompute affected string**  
+  For every name $v$ such that there exists a directed path $v\to x$
+in $G$ for some $x \in \text{FixSet}$, recompute
+```currentCandidate(v)``` in topological order.
```
jesyspa: It isn't necessary to recompute this: candidates are symbolic, and rely on the names.  You can use a memoizing system to keep most of the order-related stuff implicit.

jesyspa (commented):
Some questions

jesyspa (approved):

jesyspa (approved):

PR #3, opened 2025-12-11
Introduce ssa-conversion.md

AlexFalter:
This PR introduces a design document explaining the SSA conversion that will be implemented to convert pure Kotlin functions into Viper functions


Design/ssa-conversion.md
```diff
@@ -0,0 +1,53 @@
...
+translates to:
+```viper
+let x_0 == (0) in
+let x_1 == (x_0 + 2) in
+let x_2 == (x_0 + 3)
+let y_0 == ((x_0 == 1) ? x_1 : x_2) in
+y_0
+```
+Implementation note: To perform this translation two things are required:
+- A CFG topology (already given by the ExpEmbedding structure)
```
jesyspa: This is not true: the ExpEmbedding structure gives an AST, the CFG is implicit in this, but not directly available.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,53 @@
...
+```
+Note: An optimized encoding of the above example could be created by assigning 1 + 1 directly to x_0. In general, if no usages occur between two reassignments of the same variable we can omit one assignment.
+
+## What about branching?
+
+To convert an arbitrary program into SSA form, we will adhere to the following conversion algorithm:
+
+https://dl.acm.org/doi/10.1145/115372.115320
+
+The key idea is the following: Whenever a join in the control flow graph (CFG) of the program is encounterd and if it is necessary, a $\Phi$  function is assigned to the next 'version' of a variable. This $\Phi$  function selects the correct upstream (upstream in the CFG that is) value of the variable based on the execution path taken to reach that block. The efficient identification of necessary locations for $\Phi$ functions is determined using dominance frontiers, ensuring they are only inserted where distinct definitions actually collide. The dominance frontier of a node d is the set of nodes that are not strictly dominated by d, but have an immediate predecessor that is. A conversion may look like the following:
```
jesyspa: nit: please line wrap your md files at 80 characters or so, this makes it much easier to read the plaintext and leave comments.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,53 @@
+# Converting an ExpEmbedding into SSA form
+
+In order to translate pure Kotlin functions into Viper functions their body must be translated into a single expression. Therefore, reassignments of variables may not happen on the Viper level which is why the ExpEmbedding must be translated into SSA form.
+
+## Converting a Block to SSA form
+
+Assuming we have a block embedding with no branching, converting into SSA form is quite straight-forward:
```
jesyspa: This kind of presentation of the algorithm is easy to write, but makes a lot implicit.  Typically, it is better to be clear about what state you are maintaining and how you update it.  This also makes you think about where that state is being maintained, e.g. whether you are updating the expression or maintaining something extra.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,53 @@
+# Converting an ExpEmbedding into SSA form
+
+In order to translate pure Kotlin functions into Viper functions their body must be translated into a single expression. Therefore, reassignments of variables may not happen on the Viper level which is why the ExpEmbedding must be translated into SSA form.
+
+## Converting a Block to SSA form
```
jesyspa: Keep in mind that things are not necessarily quite so simple.  For example, an addition may have blocks on both sides.  (In general: assignments may appear as subexpressions.)  It would be good to think about how this should fit together.

Design/ssa-conversion.md
jesyspa: It is not clear to me, from this paragraph, how this really works.  Perhaps you can elaborate on what the dominance frontier looks like for some interesting case in the example below?

jesyspa (changes requested):

Design/ssa-conversion.md
jesyspa, line 1: General comment: when updating writing, it's worth fitting the added pieces into the general structure, rather than just appending them to the end.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+not happen on the Viper level which is why the ExpEmbedding must be translated
+into SSA form.
+
+## Converting a Block to SSA form
+
+Assuming we have a block embedding with no branching (branching in the CFG that
+is), converting into SSA form is quite straight-forward:
+- Traverse the ExpressionEmbeddings in the block in order
+- As soon as a reassignment is found introduce a new variable and assign it the
+  RHS of the reassignment.
```
jesyspa: "assign it to the RHS" - do you mean assign the RHS to it?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+- In subsequent usages of the 'original' variable use the newly introduced
```
jesyspa: Here and in the step above you are conflating two things that are going on: what the code does at runtime (or verification time), and what the compiler is doing.  The assignment happens within the code, while the replacement happens at compile-time.  It's important to keep these clearly separated, since it is quite hard to read otherwise.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+```
+translates to:
+```viper
+let x_0 == (1) in
+let x_1 == (x_0 + 1) in
+x_1
+```
+Note: An optimized encoding of the above example could be created by assigning
+1 + 1 directly to x_0. In general, if no usages occur between two reassignments
+of the same variable we can omit one assignment.
```
jesyspa: This is mixing up two different things:
- If an assigned value is never read, you can drop the assignment.
- If a variable is only used once, then you can inline its definition.

In SSA form, you see this as zero or one usage of a variable respectively.  However, I would not worry about it for now.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+
+To implement the above algorithm we need to introduce some state tracking the
+assignments of a variable: Then when encountering a variable usage in our block
```
jesyspa: I don't think this is really true.  What you can track is the SSA-variable that a source-variable is represented by at a particular moment.  The question here is what state to maintain to track that.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+we have two options:
+
+1. If the variable is a LHS of an assignment update the variable state to carry
+   the RHS as the new 'latest' version of the variable
```
jesyspa: I don't see how this could work; this sounds like what you would do to interpret the code, being a variable to value mapping.  However, the substitution you're doing is a renaming; knowing the binding is not sufficient (though of course you need it *somewhere* to emit it later).

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+The key idea is the following: Whenever a join in the control flow graph (CFG)
+of the program is encountered and if it is necessary, a $\Phi$ function is
+assigned to the next 'version' of a variable. This $\Phi$ function selects the
+correct upstream (upstream in the CFG that is) value of the variable based on
+the execution path taken to reach that block. The efficient identification of
+necessary locations for $\Phi$ functions is determined using dominance frontiers,
+ensuring they are only inserted where distinct definitions actually collide. The
+dominance frontier of a node d is the set of nodes that are not strictly
+dominated by d, but have an immediate predecessor that is. A node $d$ strictly
+dominates another node $d'$ if all paths to $d'$ in the CFG go through $d$.
```
jesyspa: Definitions should generally be given before the term is used; here, the order is the opposite.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+ExpEmbeddings continue to contain the original variable value and we inject the
+'latest' SSA version upon translation.
+
+## What about branching?
+
+To convert an arbitrary program into SSA form, we will adhere to the following conversion algorithm:
+
+https://dl.acm.org/doi/10.1145/115372.115320
+
+The key idea is the following: Whenever a join in the control flow graph (CFG)
```
jesyspa: From this paragraph it is again unclear to me what state you track.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+Therefore, a $\Phi$ selecting the correct upstream value must be inserted at node
+$D$. As the correct upstream value to be selected is determined by the result of
+an if-statement a ternary operator will suffice as the $\Phi$ function here
+
+Overall, this program translates to:
+```viper
+let x_0 == (0) in
+let x_1 == (x_0 + 2) in
+let x_2 == (x_0 + 3) in
+let y_0 == ((x_0 == 1) ? x_1 : x_2) in
```
jesyspa, line 93: It is not clear how you got the `x_0 == 1` here.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+![](images/SSA-CFG-example.jpg)
+
+Let's analyse the dominance frontiers $DF$ in this graph:
+
+- $DF(A)$ = $\emptyset$ - as A strictly dominates every other node
+- $DF(B)$ = $\{D\}$ - as you can reach D via C
+- $DF(C)$ = $\{D\}$ - as you can reach D via B
+- $DF(D)$ = $\emptyset$ - as D does not dominate any nodes (as it has no child)
+
+Therefore, a $\Phi$ selecting the correct upstream value must be inserted at node
```
jesyspa: It is not clear how you get this information from the dominance frontiers.  More generally, there are two reasonable approaches on how to include an algorithm in a design doc:
- Explain the algorithm, describing the entities you work with and how you compute them.
- Refer to the paper, only describing how the algorithm is reflected in the context of the project.

You can go for either one, but the middle road tends to give the benefits of neither.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,101 @@
...
+let x_0 == (0) in
+let x_1 == (x_0 + 2) in
+let x_2 == (x_0 + 3) in
+let y_0 == ((x_0 == 1) ? x_1 : x_2) in
+y_0
+```
+
+Implementation note: To perform this translation two things are required:
+- A CFG topology (can be derived from the ExpEmbedding structure)
+- A map between variable definitions and the blocks they occur in
```
jesyspa: I don't understand what you mean here.

jesyspa (changes requested):

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+avoid by carrying enough state into child nodes. 
+
+To explain our adaption of this algorithm further, we will follow the structure
+in the paper. First we will explain how to translate a single block into SSA
+form and then elaborate on how we can extend this to an arbitrary program.
+
+## Converting a Block to SSA form
+
+The only condition in the above paper for translating a basic block into SSA form
+is to traverse the expressions in the block in program order. As the embedding
```
jesyspa: I think that the assumption that everything happens within the order of a block is a bit strong.  Consider the following example:
```kotlin
run { x = 0; 2 } + run { x = 1; 2 }`
```
This translates to an ExpEmbedding where there are two blocks on the sides of a `+`.  This is still sequential control flow, but it is not clear from your description whether this will work.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+representation of a block in SnaKt already does this upon calling .toViper()
+we can perform the necessary transformations at this stage. Namely we will:
+- Keep a mapping from each source variable to its current defining expression
+in the SSAConverter. This mapping is on a block-basis (important later)
+- When a variable is read, we pass control flow to the PureLinearizer, look
```
jesyspa: This is quite hard to read; you are on the one hand personifying the parts of the program, but on the other hand still describing it as "us" doing all the work.

A good framework to think about these things is as follows:
- What needs to be done?
- What are the responsibilities of the components/classes?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+up the curent defining expression of the variable and substitute. To substitute
+a variable we will pass control to the Linearizer upon traversing a VariableEmbedding,
```
jesyspa: Is this duplicate with the part about the PureLinearizer above?  I don't quite understand what you mean.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+form and then elaborate on how we can extend this to an arbitrary program.
+
+## Converting a Block to SSA form
+
+The only condition in the above paper for translating a basic block into SSA form
+is to traverse the expressions in the block in program order. As the embedding
+representation of a block in SnaKt already does this upon calling .toViper()
+we can perform the necessary transformations at this stage. Namely we will:
+- Keep a mapping from each source variable to its current defining expression
+in the SSAConverter. This mapping is on a block-basis (important later)
```
jesyspa: Is the SSAConverter conceptually separate from the (Pure)Linearizer?  Since you're explaining the principle of how the algorithm works, you should explain to the reader what the components of your design are and then refer to them; as it is, to understand this someone first needs to read the code.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+- When a variable is read, we pass control flow to the PureLinearizer, look
+up the curent defining expression of the variable and substitute. To substitute
```
jesyspa: Hang on, but if you substitute the defining expression, are you not inlining the variable?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+## What about branching?
+
+It is important to note, that in the current version of this SSA transformation
+we assume that our program does not contain any back-edges in its CFG. When
+encountering a source variable that has no defining expression in the current
+block we can therefore safely assume, that the full defining expression must 
+exist in some previously traversed block. Therefore, introducing the following
+state in the SSAConverter will suffice to resolve any variable usage to a 
+defining expression:
+- A mapping from a block to their predecessors
```
jesyspa: It is not clear to me how this is tracked.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,136 @@
...
+function safeDivide(x: Int, y: Int): Int {
+  let res_0 == (0) in
+  let res_1 == (x / y) in
+  let res_2 == (-1) in
+  let res_3 == ((y != 0) ? res_1 : res_2) in
+    res_3
+}
+```
+This code will not verify because y might be 0, and Viper will attempt to verify the division x / y
+ unconditionally. Therefore, for partial operations like division, we must lift the ternary check
```
jesyspa, line 135: How do you know that an operation is partial?

jesyspa (changes requested):

Design/ssa-conversion.md
AlexFalter: I believe the BinaryOperatorEmbedding generated by this code does first translate the left expression of the + and then the right expression, so this case should be fine.
In general we could treat cases, where we first evaluate a later part of a block before evaluating a previous part in a similar way as it's done for unsealed blocks in the paper (section 2.3). That is if we are not sure we can evaluate this operator properly now, we insert an incomplete phi and get around to it later. Conceptually, this would just be multiple basic blocks with one unevaluated predecessor.
However, I do believe in the scope of my work no such case should appear and we can refrain from implementing this for now (at least I could not find a breaking example).

Design/ssa-conversion.md
```diff
@@ -28,17 +28,19 @@ form and then elaborate on how we can extend this to an arbitrary program.
...
-a variable we will pass control to the Linearizer upon traversing a VariableEmbedding,
-which then resolves the embedding to the name of the variable holding the current
-defining expression of the corresponding source variable.
-- When a variable assignment is traversed the PureLinearizer will call
-the SSAConverter to update the current defining expression of that source
-variable to hold the RHS of the assignment
+we can perform the necessary transformations at this stage. There are two components
+responsible for this. The SSAConverter holds state and provides functionality to perform
+an SSA transformation. The PureLinearizer, which exists to linearize an ExpEmbedding into
+a Viper expression, will call out to the SSAConverter if necessary. The following will be
```
jesyspa: This is too abstract to be helpful to the reader.  It sounds like the SSAConverter takes an AST, but that's not true.  What are the specific responsibilities?  What is the topology?

Design/ssa-conversion.md
```diff
@@ -28,17 +28,19 @@ form and then elaborate on how we can extend this to an arbitrary program.
...
+introduced:
```
jesyspa: I feel like you are still trying to present this as a summary of the changes your PR will have.  This is not very helpful to the reader: the goal is not to explain what work you will do, but how you will solve the problem.

Design/ssa-conversion.md
```diff
@@ -28,17 +28,19 @@ form and then elaborate on how we can extend this to an arbitrary program.
...
+- A mapping from source variables to a list of new variable definitions, where each variable 
+definition represents a reassignment of the original source variable. This mapping is on a
+per-block basis (important later).
```
jesyspa: Who tracks this?

Design/ssa-conversion.md
```diff
@@ -28,17 +28,19 @@ form and then elaborate on how we can extend this to an arbitrary program.
...
+- When reading a variable, the PureLinearizer will resolve the variable name to the variable
```
jesyspa: A list is best understood if the individual points are comparable, but here you are combining state and behaviour in a single list.

Design/ssa-conversion.md
```diff
@@ -62,9 +64,12 @@ block we can therefore safely assume, that the full defining expression must
 exist in some previously traversed block. Therefore, introducing the following
 state in the SSAConverter will suffice to resolve any variable usage to a 
 defining expression:
-- A mapping from a block to their predecessors
+- A mapping from a block to their predecessors. This mapping is established when
+we encounter an instruction that branches from one block into another. To achieve
+this, we need to call out to the Linearizer when encountering such instructions.
```
jesyspa: Here and above, "call out" does not actually clarify things.  The important part is to understand what problem needs to be solved, what steps need to be taken to solve it, and then explain how you spread the responsibilities for that over the different components.

Design/ssa-conversion.md
jesyspa: This is linear code; the point is that the two blocks are sequenced in an unexpected way, with an assignment happening at something other than the top level.  (Really, even without the + you still get a block that is not obviously scheduled before or after its containing block, kind of by definition of containment.)

jesyspa (changes requested):
Some general thoughts:
1. I think we are going somewhat in circles, and that you are trying to fix surface-level issues with the description, when the problem is that the design is still up in the air.  The rules for maintaining the block state are not clear to me, and for now it feels like the only part that is described in detail is what is already implemented.
2. While I would not recommend writing such a document with AI, my experience is that asking an LLM to ask you some questions about it can be quite helpful.  Here is [an example of doing this with Claude](https://claude.ai/share/ca1cddd5-3640-46c6-83a7-8b569f89a646).  Not everything it says is any good, so use your judgement as to what you should address, but it will call you out on things when they are vague.
3. The term `block` has a meaning in SnaKt (a particular kind of ExpEmbedding), and a different meaning in literature that talks about an SSA representation.  Do you always mean the former?  If not, how do you distinguish which one you mean?

Design/ssa-conversion.md
```diff
@@ -15,32 +15,34 @@ a program into SSA form from its AST:
...
+In the following explanations we will distinct between two types of blocks:
+1. The SnaKt block, where we mean the block ExpEmbedding in SnaKt containing
+a sequence of embeddings
+2. The AST block, where we mean the basic block in a program containing a sequence
+of statements adhering to the definition in the paper (and in literature in general)
+
+## Translating an AST block into SSA form
+
+Adapting the approach from the paper we will do the following while traversing
+an AST block:
```
jesyspa, line 33: Hang on, if AST blocks are a different thing, how do you obtain them?

Design/ssa-conversion.md
```diff
@@ -15,32 +15,34 @@ a program into SSA form from its AST:
...
+
+- Maintain a mapping from source variables to a list of defining expressions:
```
jesyspa, line 35: Won't this give the wrong order?  Consider:
x -> [1, y]
y -> [x]
How do you know whether we did `x = 1; y = x; x = y` or `x = 1; x = y; y = x`?

Design/ssa-conversion.md
```diff
@@ -55,31 +57,23 @@ let x_1 == (x_0 + 1) in
...
-When reading a variable, that has no defining expression in the current basic
-block, we do the following:
-- If the basic block only has one predecessor, use the current definition
-of the source variable of that block
-- If there is more than one predecessor, use the conditional information tracked
-to construct a $\Phi$ function selecting the correct upstream value of the variable.
-use this $\Phi$ function as the current defining expression of the variable.
-
-Overall this approach will result in the following translation behaviour:
+### Implmentation details
```
jesyspa, line 60: nit: Implementation

Design/ssa-conversion.md
```diff
@@ -55,31 +57,23 @@ let x_1 == (x_0 + 1) in
...
+
+The problem we are facing while employing the algorithm from the paper in SnaKt
+is that SnaKt blocks differ from AST blocks. We can get around this by doing
+the following:
+
+Whenever we traverse an embedding, that branches into another AST block, we make
+note of the topology of the AST (that is what new blocks exist and under what
+conditions we branch into those). We then translate each of those AST blocks as
+described above and as described in the paper (inserting $\Phi$ functions when
+joining two AST blocks and if necessary).
```
jesyspa, line 74: To me this paragraph reads as "we just do the thing that makes things work", which does not specify the design.

jesyspa (changes requested):
I think that this version gives a better presentation of the parts that are already clear, but I do not think it answers any further questions.  Please take a few days to think about the design, work out examples on paper (including an overview of what information you track), and then formulate the algorithm as a whole.

PR #4, opened 2025-12-18
Introduce SSA conversion

AlexFalter:
This PR introduces ssa-conversion.md in the design files. This file describes methodology to apply an algorithm to SnaKt's embedding structure to convert a program into SSA form


Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+
+When converting a pure Kotlin function into a Viper function, we have to translate
+the program into a single expression. This is realised in a chain of let bindings.
+In those, variable reassignments may not occur. We therefore need to translate
+the code into SSA form to avoid any reassigning that might happen.
+
+## Theoretical background
+
+The most common algorithm for translating a program into SSA form is described
+in a paper by Cryton et al.
```
jesyspa: typo: Cyrton

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+the program into a single expression. This is realised in a chain of let bindings.
+In those, variable reassignments may not occur. We therefore need to translate
+the code into SSA form to avoid any reassigning that might happen.
+
+## Theoretical background
+
+The most common algorithm for translating a program into SSA form is described
+in a paper by Cryton et al. 
+
+https://www.cs.utexas.edu/~pingali/CS380C/2010/papers/ssaCytron.pdf
```
jesyspa: Please provide citations for papers; you can include a link, but the title + authors are still key.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+While at first glance those expressions and their side-effect in current scope
+may cause issues, we can check that as long as we traverse the expression in 
+the program order it is being evaluated in, we will construct the assignments
+one after the other and use the correct definition accordingly.
+
+## SnaKt block to AST block
+
+To create a topology in the paper's sense the following state will be introduced
```
jesyspa: Which paper uses the term topology, and what does it mean?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+
+https://c9x.me/compile/bib/braun13cc.pdf
+
+## SnaKt block vs. AST block
+
+In the AST representation used in the paper a block is a sequence of statements.
+That is a block only contains linear code. This is not true in SnaKt. In SnaKt
+a block is a sequence of ExpEmbeddings, which contain - among other things - also
+branching embeddings. The main challenge we have to overcome therefore is to
+use the SnaKt representation of a block as the paper's representation of a block.
```
jesyspa: But you do not do this.  Instead, what you seem to have is an implicit representation of AST blocks by using the traversal of the ExpEmbedding.  This is fine, it's probably the right way to go, but it is completely different from what you say here.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Putting it all together
+
+After the whole embedding is linearized we have a topology of SSAConverters
+holding variable assignments in their scope. From this topology we can
+construct the let chain by traversing the SSAConverters from the top-level
+SSAConverter in dependency order (that is only traverse into one SSAConverter
+if all its predecessors have already been traversed into). During this traversal
+we can construct an ordered list of assignments, which we can then translate into
+a let chain. A few cases need considering:
```
jesyspa: This looks unfinished.  I guess you mean that the subsections below are the cases, but using a colon like this gives the wrong impression.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+most recent as the paper does. 
+
+Looking at the implementation we will introduce the following:
+- The SSAConverter maintains a list of assignments. An assignment is a mapping
+between source variable and defining expression
+- The last expression in this list mapping a source variable to a defining expression
+represents the most recent defining expression we encountered for that source variable, 
+the one before-hand the 2nd most recent and so on. 
+- When encountering a variable read, we substitute a variable holding the most
+recent definition for that source variable into the expression
```
jesyspa: If I am reading the above correctly, you are storing a list of (Variable, Expression) pairs.  Here you want to perform a Variable -> Variable substitution.  How do you do this?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+
+    for every subblock in subblocks:
+        translate(linearizer: Linearizer(ssaConverter: newConverters[subblock]))
+```
+*(\*)* A subblock is any block the embedding has branches into (e.g. an if-embedding
+has two subblocks - the then block and else block)
+
+The resulting topology between SSAConverters is similar to the topology between
+AST blocks. We can therefore directly use the global value numbering algorithm
+described in the paper on this construct to convert into SSA form.
```
jesyspa: As discussed in Slack, please describe the algorithm.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+    subblocks     <- list of subblocks in embedding (*)
+    joinConverter <- SSAConverter(predecessors: {}, successors: this.ssaConverter.successors)
+    newConverters <- {}
+
+    for every subblock in subblocks:
+        newConverters.add(SSAConverter(predecessors: this.ssaConverter, successors: joinConverter))
+
+    joinConverter.predecessors <- newConverters
+    this.ssaConverter.successors <- newConverters
+    this.ssaConverter <- joinConverter
```
jesyspa: This is a very low-level description.  Are you just saying that you create a converter per subblock and an extra synthetic one for the join block, and then link them in a particular order?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+SSAConverter in dependency order (that is only traverse into one SSAConverter
+if all its predecessors have already been traversed into). During this traversal
+we can construct an ordered list of assignments, which we can then translate into
+a let chain. A few cases need considering:
+
+### Conditions
+
+To distinct under what condition a variable assignment is valid (and to construct
+the $\Phi$ functions mentioned in the paper) the SSAConverter will additionally
+to the above also hold a 'branching predicate', which evaluates to true iff.
```
jesyspa: I don't think this is appropriate to delegate to a side issue like this.  This is part of the information you need to track, and you need to be aware of where to do this.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+the subblock represented by the SSAConverter is being branched into.
+
+### (Early) returns
+
+To handle early returns each SSAConverter holds their return expression (currently
+implemented as body, but should probably be renamed). In combination with the above
+information on when a SSAConverter is being traversed into, we can construct
+the overall return expression during the above mentioned dependency traversal by
+combining each encountered return expression into one overarching ternary expression
+that returns the expression from a subblock if their condition is met. If a SSAConverter
```
jesyspa: I don't understand what you mean by this.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,128 @@
...
+a let chain. A few cases need considering:
+
+### Conditions
+
+To distinct under what condition a variable assignment is valid (and to construct
+the $\Phi$ functions mentioned in the paper) the SSAConverter will additionally
+to the above also hold a 'branching predicate', which evaluates to true iff.
+the subblock represented by the SSAConverter is being branched into.
+
+### (Early) returns
```
jesyspa: You could look into doing this by treating return as a control flow block with no successor.

jesyspa (changes requested):
I think this is better, but it still does not really bridge the gap of how we traverse the expression and how the paper builds these things.

FYI, this is the last review I have time for this week, so please don't hurry with another version.  It can be useful to let it sit for a day or two.  You can start implementing it if you're confident of the approach.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
+# Converting an ExpEmbedding into SSA form
+
+When converting a pure Kotlin function into a Viper function, we have to translate
+the program into a single expression. This is realised in a chain of let bindings.
```
jesyspa: nit: body

jesyspa: nit: "this" is unclear; the goal is to support local variables and reassignment

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+working on AST nodes, developed by Braun et al. [2]
+
+## SnaKt block vs. AST block
+
+In the AST representation used by Braun et al. a block is a sequence of statements [2].
+That is, a block only contains linear code. This is not true in SnaKt. In SnaKt
+a block is a sequence of ExpEmbeddings, which contain - among other things - also
+branching embeddings. The main challenge we have to overcome therefore is to
+bridge the gap between the SnaKt representation of a block and the paper's representation of a block.
+In the following, we will call the SnaKt version a 'SnaKt block' and the Braun et al.
```
jesyspa: Do you need to care about SnaKt blocks at all?  I don't see a case when they're relevant.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+## Local value numbering
+
+Before converting our structure into one compatible with the paper's algorithm, 
+let us first assume that we have a single AST block we can traverse in program order
+and discuss any changes we have to make to the local value numbering algorithm [2]
+to create something similar for SnaKt.
+
+The key difference is that, as we are not swapping variables in place, but rather
+track *all* assignments to construct a linear let chain out of potentially non-linear
+code. We therefore need to maintain all assignments rather than only the
```
jesyspa: It is not clear what you mean here.  In particular:

> as we are not swapping variables in place, but rather track *all* assignments to construct a linear let chain out of potentially non-linear code

"as X, Y" is the usual form, but this sentence is missing the Y.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+to create something similar for SnaKt.
+
+The key difference is that, as we are not swapping variables in place, but rather
+track *all* assignments to construct a linear let chain out of potentially non-linear
+code. We therefore need to maintain all assignments rather than only the 
+most recent as the paper does. 
+
+We will introduce the following:
+- The SSAConverter maintains a list of assignments in its scope.
+An assignment is a mapping between source variable and defining expression
```
jesyspa: Is this true?  Your last PR had the SSAConverter working with triples.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+- The last expression in this list mapping a source variable to a defining expression
+represents the most recent defining expression we encountered for that source variable
+in the SSAConverter's scope, the one before that the 2nd most recent and so on. 
+- When encountering a variable assignment, we update the list to hold the 
+RHS of the assignment as the latest encountered defining expression. Further, we give
+this assignment a unique index. This index will be shared across SSAConverters such
+that when we are constructing the overall expression, each assignment has a unique
+per source variable index we can identify it by. In the resulting expression this
```
jesyspa: Why not track this information separately, by the way?  It seems strange to go through the whole list every time you want to resolve a variable use.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+## SnaKt block to AST block
+
+In the above we have assumed linear code consisting of a single AST block. To
+support any branching that might occur while translating an ExpEmbedding,
+we will construct the same topology AST blocks have between SSAConverters.
+To do so we will introduce the following for every SSAConverter:
+- A list of preceding SSAConverters
+- A list of succeeding SSAConverters
+
+Consider an If-Embedding being translated. The PureLinearizer will modify
```
jesyspa: This is the first reference to `PureLinearizer`.  What is the conceptual split between it and SSAConverter?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+will have any successor the current SSAConverter might have as its successor.
+Predecessors will be established accordingly.
+
+After this relationship between SSAConverters is established the delegating PureLinearizer
+will initiate the translation of both the then and else branch. After that it will update its SSAConverter
+to be the joining SSAConverter as this shall hold the variable definitions for any
+assignments occurring after the branch. Note that as the PureLinearizer is being copied
+with adjusted position information when a WithPosition-Embedding, which encapsulates
+an If-Embedding, is traversed rather than updating the reference to the SSAConverter,
+we will update an object holding the actual reference. This allows the change
```
jesyspa: I don't understand what you mean by this sentence.  Relying on there being a `WithPosition` node doesn't seem like a good idea, though.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+- One SSAConverter will be introduced representing a join node behind the
+two branches
+
+The current SSAConverter will have the two branch SSAConverters as its successors.
+These will have the joining SSAConverter as their successor and the joining converter
+will have any successor the current SSAConverter might have as its successor.
+Predecessors will be established accordingly.
+
+After this relationship between SSAConverters is established the delegating PureLinearizer
+will initiate the translation of both the then and else branch. After that it will update its SSAConverter
```
jesyspa: But the PureLinearizer generally does not initiate translation like this.  It sounds like there's some kind of conceptual change going on that's not in the text?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+to be the joining SSAConverter as this shall hold the variable definitions for any
+assignments occurring after the branch. Note that as the PureLinearizer is being copied
+with adjusted position information when a WithPosition-Embedding, which encapsulates
+an If-Embedding, is traversed rather than updating the reference to the SSAConverter,
+we will update an object holding the actual reference. This allows the change
+to be propagated to the PureLinearizer object translating any statements after the branch.
+
+For now, the implementation will restrict to the above case of an If-Embedding, only
```
jesyspa: Same here, the PureLinearizer should not know what kind of embedding it is translating.  I think there's some context missing here.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Phi expressions
+
+Consider a SSAConverter succeeding the then and else branch of an if statement.
+Further imagine, that both branches modify a variable $x$. If in the scope of this
+SSAConverter $x$ is being used, we must be able to resolve to either the name defined
+in the then branch or the name of the else branch depending on previous control flow.
+This can be achieved as follows:
+
+Each SSAConverter holds a boolean expression under which it is being traversed. This condition
```
jesyspa: Please add an example for what's described in this paragraph.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+can be extracted from the if-statement upon constructing the topology between SSAConverters.
+Note, that in the case of nested branching the innermost SSAConverter will only hold the condition
+of the innermost if, the 2nd innermost only the condition of the 2nd innermost if and so on. 
+We will call this condition local branching condition. As the created topology between
+SSAConverters ensures that there is a joining converter present for any branching,
+only storing the 'local' condition is sufficient (and makes the resulting expression cleaner).
+
+For the above described case we can now collect the variable definitions for $x$ from
```
jesyspa: It isn't clear where or when this occurs.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+    # Sole predecessor must hold the corresponding definition
+    if this SSAConverter only has one predecessor pre:
+        return resolveVariableName(pre)
+
+    # Multiple predecessors have definitions
+    else:
+        incomingNames := resolveVariableName(toResolve) for all predecessors
+        phiExpression := phi expression selecting the correct name from all incoming names
+        add an assignment of phiExpression to toResolve in this converter
+        return the name holding this assignment
```
jesyspa: It's not obvious to me this will always be the case.  E.g.

```python
x = 3
if c:
  y = 0
else:
  y = 1
return x+y
```

`x` does not need a φ expression, but that block has multiple predecessors.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+## (Early) returns
+In our function we may have multiple return statements at any branch. We need to
+construct a single return expression from these returns. Note, that the
+SSAConverter in scope of the branch of the return expression will hold this expression as its body.
+To construct the overall expression, we will consider the global branching condition
```
jesyspa: This is the first time you mention a global branching condition.  What is the state you are tracking, and where is it tracked?

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+a block is a sequence of ExpEmbeddings, which contain - among other things - also
+branching embeddings. The main challenge we have to overcome therefore is to
+bridge the gap between the SnaKt representation of a block and the paper's representation of a block.
+In the following, we will call the SnaKt version a 'SnaKt block' and the Braun et al.
+version an 'AST block'
+
+## Local value numbering
+
+Before converting our structure into one compatible with the paper's algorithm, 
+let us first assume that we have a single AST block we can traverse in program order
```
jesyspa: I think you may be trying to handle the simple case first to make things more approachable, but I don't think it actually helps.  The goal is to communicate an algorithm.  By cutting it up into a bunch of assumptions, you are showing pieces of it without ever actually describing it as a whole, and you don't get anything out of it, since these pieces are not easier than the corresponding part of the whole.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+
+## (Early) returns
+In our function we may have multiple return statements at any branch. We need to
+construct a single return expression from these returns. Note, that the
+SSAConverter in scope of the branch of the return expression will hold this expression as its body.
+To construct the overall expression, we will consider the global branching condition
+of an SSAConverter. That is in the case of nested branching the global branching condition -
+contrary to the local branching condition - will hold the conjunction of all conditions
+that need to be satisfied for control flow to reach this statement. We can then
+construct an overall return expression by:
```
jesyspa: Is this a second traversal?  Some extra data you track?  It's not clear to me.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Putting it all together
+
+We now have a topology between SSAConverters holding variable names and their definitions
+as well as an overall return expression.
+To construct a single expression from this we can do the following:
+
+1. Create a dependency graph traversal ordering of all SSAConverters
+2. Collect all assignments in that ordering
+3. Create a single chain of let bindings, where the outermost let-binding
```
jesyspa: Why is it necessary to construct it after the fact, rather than maintaining that information as you go?

jesyspa (changes requested):

Design/ssa-conversion.md
AlexFalter: What information do you mean here exactly?

Design/ssa-conversion.md
AlexFalter: I'll adjust the phrasing: Essentially as the PureLinearizer is being copied due to WithPosition embeddings being present, I cannot simply update the reference the PureLinearizer holds by saying `this.ssaConverter = joinConverter` as the PureLinearizer is a different object to the one translating subsequent embeddings of the if-embedding, so I need to update an underlying object. The implementation uses `KProperty` for this. So it does not matter whether or not a `WithPosition` embedding is present or not

Design/ssa-conversion.md
AlexFalter: Is this with regards to the 1., 2. or 3. (or all)?

Design/ssa-conversion.md
jesyspa: Thanks, that makes more sense (and also clarifies the implementation PR somewhat).  I think this underlines that more context needs to be given: there are multiple `PureLinearizer`s and this fact is relevant for the algorithm.

Design/ssa-conversion.md
jesyspa: You have two mappings you are interested in:
- Source variable -> latest SSA variable
- SSA variable -> defining expression

You are currently storing this as a list of triples.  Why, rather than keeping a map and a list?

Design/ssa-conversion.md
jesyspa: All 3 seem like things that could be done while building the SSA converters.

AlexFalter: I've decided against it for the following reasons:

1. The dependency graph traversal is not really clear when we are building the SSAConverters. Imagine an If-Embedding resulting in a diamond shape - we might traverse another if-embedding in the then branch of the previous. This will change the previously seen dependency graph. So I decided that creating the complete traversal after the whole embedding is seen is more consistent
2. I did not really see the value of having an overarching list of assignments tracked. As I need the dependency graph for the return expressions regardless it seems natural to just collect the assignments at this point as well and save extra state before-hand. Further, each SSAConverter only tracking assignments in their own scope seemed more natural to me and other components shouldn't really track anything
3. Having a partial let-chain seems unnatural to me. It would entail introducing a mutable expression we then somewhat fill. As I need to track the assignments anyway there is also no use-case for this partial expression before we construct the overall result, so I decided against it

Design/ssa-conversion.md
AlexFalter: I think what helps here is that the list of assignments is ordered. So the latest SSA variable always sits at the end of the list and we have the first mapping implicitly given

Design/ssa-conversion.md
jesyspa: I've thought a bit about the design as a whole and it got rather long for a comment here.  I wrote up what seems reasonable to me in a doc: https://docs.google.com/document/d/1KWd1GawUPsIgHxQlyyCyEKzyi8agSoW61WUTzTWMoQI/

I think this is simpler overall, but I may be missing why this wouldn't work.  I hope it also illustrates some of the earlier stylistic feedback I had.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,158 @@
...
+
+In the AST representation used by Braun et al., a block is a sequence of
+statements [2]. That is, a block only contains linear code. This is not true
+in SnaKt. In SnaKt, a block is a sequence of `ExpEmbedding`s, which
+may contain—among other things—branching embeddings. The main challenge we
+have to overcome is bridging the gap between the SnaKt representation of a
+block and the paper's representation.
+
+We hence introduce a graph topology similar to the AST representation
+required by Braun et al. There are three types of nodes in this graph:
```
jesyspa, line 41: nit: not three

Design/ssa-conversion.md
```diff
@@ -0,0 +1,158 @@
...
+accordingly.
+
+## Literature
+[1] Cytron, R.; Ferrante, J.; Rosen, B. K.; Wegman, M. N.; Zadeck, F. K.
+(1991). [Efficiently computing static single assignment form and the control
+dependence graph](https://www.cs.utexas.edu/~pingali/CS380C/2010/papers/ssaCytron.pdf)
+
+[2] Braun, M.; Buchwald, S.; Hack, S.; Leißa, R.; Mallon, C.; Zwinkau, A.
+(2013). [Simple and Efficient Construction of Static Single Assignment Form]
+(https://c9x.me/compile/bib/braun13cc.pdf)
```
jesyspa: Please put this on one line, or the link won't render

Design/ssa-conversion.md
```diff
@@ -0,0 +1,158 @@
...
+1. Delegates to the current `SSABlockNode` to update its map from source
+   variable names to `SSAVariableName`s to hold a fresh `SSAVariableName` for
+   the source variable on the left side of the assignment.
+2. Adds an `SSAAssignment` consisting of the new `SSAVariableName` and the
+   right side of the assignment as its defining expression.
+
+### Encountering a Variable Usage
+
+When a variable is used in an expression, the `SSAConverter` looks up which
+`SSAVariableName` holds the defining expression for that source variable at
```
jesyspa: This is confusing; you aren't actually doing anything with defining expressions.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,158 @@
...
+the current point in the function body via the current `SSABlockNode`.
+
+The `SSABlockNode` finds the name as follows:
+- If it has a mapping for the source variable, that version is returned.
+- Otherwise, it looks up the name in its predecessor.
```
jesyspa, line 90: The motivation would be nice to mention.

Design/ssa-conversion.md
```diff
@@ -0,0 +1,158 @@
...
+3. Creates an `SSAJoinNode` to join both of the newly created `SSABlockNode`
+   branches.
+4. Creates another `SSABlockNode` succeeding the `SSAJoinNode` to translate
+   anything following the branch.
+
+## After the Function Body is Traversed
+
+After the whole function body is traversed, the `SSAConverter` is called to
+construct the resulting chain of `let` bindings. To do this, all collected
+return expressions are first folded into a single return expression.
```
jesyspa, line 132: This is the first time you mention return expressions; it may be worth saying a bit more about what state you track for them.

jesyspa (dismissed):
Some feedback, but overall this looks okay; please clarify the points brought up to the degree that looks necessary to you, and then you can go ahead and merge.

Design/ssa-conversion.md
AlexFalter: I'm a bit confused: Do you mean you want me to write 3 - because there is a JoinNode, BlockNode and a StartNode

Design/ssa-conversion.md
AlexFalter: The state tracked for return expressoins is mentioned on lines 59-64 - should it be more elaborate?

Design/ssa-conversion.md
jesyspa: Ah no, sorry, I just misread.  LGTM.

Design/ssa-conversion.md
jesyspa: I think adding a paragraph on "encountering a return" would make sense given the rest of the structure

jesyspa (dismissed):

jesyspa (approved):
LGTM.

PR #5, opened 2026-03-01
Heap-dependent expressions design

AlexFalter:
This PR introduces a design for heap-dependent expressions. Note, that for now no implementation details are provided. If the design is settled, implementation details will be added

### DIFF
--- a/Design/heap-dependent-expressions-in-pure-functions.md
+++ b/Design/heap-dependent-expressions-in-pure-functions.md
@@ -175,8 +175,9 @@ function isLastIndex(node: Ref, index: Int): Bool
 ### In a method
 
 Analogous to SnaKt's existing permission system, we will inhale all relevant
-permissions before the function gets called. Note, however, that we do not
-unfold them (as we require them to be present in the function).
+permissions before the function gets called and exhale them after the function
+was called. Note, however, that we do not unfold them (as we require them to be
+present in the function).
 
 ```kotlin
 fun getLastIndex(node: Node): Int {
@@ -190,6 +191,7 @@ method getLastIndex(node: Ref) returns (res: Int)
 {
     inhale acc(purePredicate(node), wildcard)
     var len: Int := length(node)
+    exhale acc(pure_list(node), wildcard)
     res := len - 1
 }
 ```
--- a/Design/heap-dependent-expressions-in-pure-functions.md
+++ b/Design/heap-dependent-expressions-in-pure-functions.md
@@ -1,0 +1,220 @@
+# Heap-dependent expressions in pure functions
+
+For a pure function, heap-dependent expressions do not violate purity per se.
+In general, as long as we only read from the heap and can ensure that no
+aliasing references exist, the function body remains pure. To add support for
+these function bodies, we will introduce a new predicate granting permissions
+to all relevant fields used within them. We will require this predicate for
+all encountered reference types in the function's precondition and unfold it
+to allow the accesses made within the function. Any functions violating the
+purity assumption will be rejected by the purity checker.
+
+## The pure predicate
+
+The pure predicate is nearly equivalent to the shared predicate, with the only
+difference being that it also acquires permissions on mutable fields. Consider
+the following example:
+
+```kotlin
+data class Node(val next: Ref, var value: Int)
+```
+This class has the corresponding pure predicate:
+```viper
+predicate purePredicate(this: Ref) {
+    acc(this.next, wildcard) &&
+    acc(this.value, wildcard) &&
+    (this.next != null ==> acc(purePredicate(this.next), wildcard))
+}
+```
+
+## Constructing the precondition
+
+Note that all reference types for which we need to require the pure predicate
+can be found in the function's parameters. Furthermore, because the predicate
+also contains the predicates for any references within the parameters, no extra
+predicates need to be acquired to ensure that potentially nested heap accesses
+in the function body work. However, the predicate only contains the predicates
+of nested references if they are non-null (see the predicate definition above).
+We will discuss the implications of this constraint in the "Nested field
+accesses" section.
+
+## Translation
+
+We will unfold all necessary predicates before an access. As an optimization,
+we want to keep track of previously unfolded predicates to avoid unfolding them
+again (although it is possible to do so). This ensures that there are
+sufficient permissions for all accesses made in the function. However, this
+approach comes with some limitations. To elaborate further, consider the
+following examples:
+
+### Simple field accesses
+
+For this simple case, we can clearly see that the approach yields a working
+function.
+
+```kotlin
+@Pure
+fun getFirstValue(node: Node): Int {
+    return node.value
+}
+```
+translates to:
+```viper
+function getValue(node: Ref): Int
+    requires acc(purePredicate(node), wildcard)
+{
+    unfolding purePredicate(node) in
+        node.value
+}
+```
+
+### Nested field accesses
+
+In the example below, we must also unfold the predicate of the reference to
+`next` to ensure the access made to the next node is valid. In general, nested
+field accesses can be understood as a path. When encountering a nested field
+access, we need to unfold all predicates for the types on the path that have
+not been previously unfolded. A natural consideration is to simply unfold as
+much as possible at the beginning of the function. This, however, would not
+work for the example below, as unfolding the predicates implicitly assumes that
+the path does not contain a null reference. Below, this is only true in the
+`else` branch of the ternary operator (or the Elvis operator in the Kotlin
+version); hence, an on-demand solution is required.
+
+```kotlin
+@Pure
+fun getThisOrNextValue(node: Node): Int {
+    return node.next?.value ?: node.value
+}
+```
+
+```Viper
+function getThisOrNextValue(node: Ref): Int 
+    requires acc(purePredicate(node), wildcard)
+{
+    unfolding purePredicate(node) in 
+        node.next == null ? 
+            node.val : 
+            unfolding purePredicate(node.next) in node.next.val
+}
+```
+
+### Function calls to self
+
+Even though we are unfolding the predicate required by the called function,
+this translation still works. This is because, in a verification context,
+unfolding a predicate with wildcard permissions does not cause that context to
+lose wildcard permissions to the predicate. 
+
+```kotlin
+@Pure
+fun length(node: Node): Int {
+    return if (node.next == null) {
+        1
+    } else {
+        length(node.next)
+    }
+}
+```
+
+```Viper
+function length(node: Ref): Int 
+    requires acc(purePredicate(node), wildcard)
+{
+    unfolding purePredicate(node) in 
+        node.next == null ? 
+            1 : 
+            length(node.next)
+}
+```
+
+### Limitations
+
+Note that we are acquiring more permissions than necessary. We solely infer the
+required predicates from the function's parameters. While this approach significantly
+simplifies the problem, it has some limitations. An access to a class might only occur
+under certain conditions in the function, or not at all. We still require permissions for
+the pure predicate in those cases, which could potentially cause verification
+to fail even though the function is valid.
+
+## When the function gets used
+
+There are three possible places where a function may be used: in another
+function, in a method, or in a specification. We will detail these use cases
+below.
+
+### In another function
+
+In this case, we need to unfold predicates in a similar way as we did for
+field accesses before calling the function. Note that any pure predicates
+required by the called function must either be directly available in the
+caller's precondition or be contained within the predicates in the caller's
+precondition.
+
+```kotlin
+@Pure
+fun isLastIndex(node: Node, index: Int): Bool {
+    return if (length(node) == index + 1) {
+        true 
+    } else {
+        false
+    }
+}
+```
+
+```Viper
+function isLastIndex(node: Ref, index: Int): Bool 
+    requires acc(purePredicate(node), wildcard)
+{
+    unfolding purePredicate(node) in 
+        length(node) == index + 1 ? 
+            true : false
+}
+```
+
+### In a method
+
+Analogous to SnaKt's existing permission system, we will inhale all relevant
+permissions before the function gets called. Note, however, that we do not
+unfold them (as we require them to be present in the function).
+
+```kotlin
+fun getLastIndex(node: Node): Int {
+    val len = length(node)
+    return len - 1
+}
+```
+
+```Viper
+method getLastIndex(node: Ref) returns (res: Int)
+{
+    inhale acc(purePredicate(node), wildcard)
+    var len: Int := length(node)
+    res := len - 1
+}
+```
+
+### In a specification
+
+If a function is used in a specification, that context must also require the
+predicates from the function's precondition to ensure the specification is
+valid.
+
+```kotlin
+fun getFirstValue(node: Node): Int {
+    preconditions {
+        length(node) > 0
+    }
+    return node.value
+}
+```
+
+```Viper
+method getFirstValue(node: Ref) returns (res: Int)
+    requires acc(purePredicate(node), wildcard)
+    requires length(node) > 0
+{
+    inhale acc(sharedPredicate(node), wildcard)
+    res := node.value
+}
+```
--- a/Design/heap-dependent-expressions-in-pure-functions.md
+++ b/Design/heap-dependent-expressions-in-pure-functions.md
@@ -40,12 +40,26 @@ accesses" section.
 
 ## Translation
 
-We will unfold all necessary predicates before an access. As an optimization,
-we want to keep track of previously unfolded predicates to avoid unfolding them
-again (although it is possible to do so). This ensures that there are
-sufficient permissions for all accesses made in the function. However, this
-approach comes with some limitations. To elaborate further, consider the
-following examples:
+As there is only a single expression in a pure function, we cannot simply unfold
+provided predicates from the precondition but rather need to place unfolding in
+expressions in the Viper encoding to ensure accesses are valid. We introduce a
+hybrid approach between greedily unfolding at the beginning of a function and 
+unfolding on-demand upon encountering a field access. Namely in SnaKt, after
+traversing a function body we have a list of `SSAAssignment`s as well as an overarching
+return expression. In the current encoding as a 'let chain' branch results are written
+to a dedicated variable and later another intermediate variable is introduced to
+resolve between the two branch results. Notice, that the branch results are defined
+regardless of whether or not the branch is used. Hence, we can unfold any predicate
+used in the let assignments at the beginning of the function as there will be no
+further condition that must be met in the current encoding for these predicates
+to be unfolded. Limitations of this approach will be discussed later. 
+For return expressions this is different. In the current encoding, the expression
+of the innermost let-binding is a potentially nested Ternary expression.
+Upon encountering a field access on a variable from the let-chain
+in said return expression, we unfold any further predicates necessary. Note, that these
+might be predicates already unfolded previously - hence as an optimization we might want
+to keep track of what predicates are unfolded at the beginning of the function to
+catch such cases.
 
 ### Simple field accesses
 
@@ -136,6 +150,42 @@ simplifies the problem, it has some limitations. An access to a class might only
 under certain conditions in the function, or not at all. We still require permissions for
 the pure predicate in those cases, which could potentially cause verification
 to fail even though the function is valid.
+
+Further, there is a limitation in verification capibilities arising from the fact
+that conditional assignments write results to intermediate values regardless
+of the conditoin that must be met for these to hold. Consider the following
+example:
+
+@Pure
+fun getNextValueOrDefault(node: Node): Int {
+    val defaultOrValue = if (node.next == null) {
+        -1
+    } else {
+        node.next.value
+    }
+    return defaultOrValue
+}
+```
+
+This will resolve to:
+
+```Viper
+function getNextValueOrDefault(node: Ref): Int 
+    requires acc(purePredicate(node), wildcard)
+{
+    unfolding purePredicate(node) in
+    unfolding purePredicate(node.next) in
+    let anon1 == (-1) in
+    let anon2 == (node.next.value) in
+    let anon3 == ((node.next == null) ? anon1 : anon2) in
+        anon3
+}
+```
+
+Clearly, this won't verify due to insufficient permissions to the purePredicate
+of node.next. Hence, this is a limitation of this approach. However note, that 
+the concious user can avoid this by boxing nullity checks into the return expression.
+For an example of this please note the example in 'Nested Field Accesses' above
 
 ## When the function gets used
 
--- a/Design/heap-dependent-expressions-in-pure-functions.md
+++ b/Design/heap-dependent-expressions-in-pure-functions.md
@@ -3,7 +3,8 @@
 In this document, we will discuss the design and implementation of heap-dependent
  expressions in pure functions. We will take a look at the Viper encoding prior to 
  this change, how to extend it to support heap-dependent expressions, and finally, 
- what limitations arise from the designed encoding.
+ what limitations arise from the designed encoding. More detailed information
+ can be found in the Bachelor's thesis of Alexander Falter.
 
 ## Viper encoding prior and problem definition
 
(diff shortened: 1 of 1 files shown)


Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
+# Heap-dependent expressions in pure functions
+
+For a pure function, heap-dependent expressions do not violate purity per se.
+In general, as long as we only read from the heap and can ensure that no
+aliasing references exist, the function body remains pure. To add support for
```
jesyspa: I don't see how aliasing references are relevant here, could you elaborate?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+purity assumption will be rejected by the purity checker.
+
+## The pure predicate
+
+The pure predicate is nearly equivalent to the shared predicate, with the only
+difference being that it also acquires permissions on mutable fields. Consider
+the following example:
+
+```kotlin
+data class Node(val next: Ref, var value: Int)
```
jesyspa: Is the type of `next` correct?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+in the function body work. However, the predicate only contains the predicates
+of nested references if they are non-null (see the predicate definition above).
+We will discuss the implications of this constraint in the "Nested field
+accesses" section.
+
+## Translation
+
+We will unfold all necessary predicates before an access. As an optimization,
+we want to keep track of previously unfolded predicates to avoid unfolding them
+again (although it is possible to do so). This ensures that there are
```
jesyspa: This sounds like you're using the `unfold` statement, but that's not available in a pure setting.  Are you going to do `unfolding ... in` to ensure that every access is within that `in`?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+    acc(this.next, wildcard) &&
+    acc(this.value, wildcard) &&
+    (this.next != null ==> acc(purePredicate(this.next), wildcard))
+}
+```
+
+## Constructing the precondition
+
+Note that all reference types for which we need to require the pure predicate
+can be found in the function's parameters. Furthermore, because the predicate
```
jesyspa: What are your thoughts on global variables?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+### Limitations
+
+Note that we are acquiring more permissions than necessary. We solely infer the
+required predicates from the function's parameters. While this approach significantly
+simplifies the problem, it has some limitations. An access to a class might only occur
+under certain conditions in the function, or not at all. We still require permissions for
+the pure predicate in those cases, which could potentially cause verification
+to fail even though the function is valid.
```
jesyspa: I'm not sure I follow - this sounds like a limitation of how you obtain the pure predicate.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+precondition.
+
+```kotlin
+@Pure
+fun isLastIndex(node: Node, index: Int): Bool {
+    return if (length(node) == index + 1) {
+        true 
+    } else {
+        false
+    }
```
jesyspa: Unless it's needed for the example, consider `return lenth(node) == index + 1`.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+    return len - 1
+}
+```
+
+```Viper
+method getLastIndex(node: Ref) returns (res: Int)
+{
+    inhale acc(purePredicate(node), wildcard)
+    var len: Int := length(node)
+    exhale acc(pure_list(node), wildcard)
```
jesyspa: This does not work for exhaling wildcard permissions, since you still have access to them.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+fun getLastIndex(node: Node): Int {
+    val len = length(node)
+    return len - 1
+}
+```
+
+```Viper
+method getLastIndex(node: Ref) returns (res: Int)
+{
+    inhale acc(purePredicate(node), wildcard)
```
jesyspa: This is inhaling information you may already partially have, namely whatever is in the shared predicate.  It would be nice to be able to use the information about the values of those fields, if it is available.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,222 @@
...
+    var len: Int := length(node)
+    exhale acc(pure_list(node), wildcard)
+    res := len - 1
+}
+```
+
+### In a specification
+
+If a function is used in a specification, that context must also require the
```
jesyspa: This is a major change, which you only briefly mention.  How do you plan to implement this?

jesyspa (changes requested):

Design/heap-dependent-expressions-in-pure-functions.md
AlexFalter: I just noticed that they are not - I'll remove it

Design/heap-dependent-expressions-in-pure-functions.md
AlexFalter: Nope :))

Design/heap-dependent-expressions-in-pure-functions.md
AlexFalter: Discussed over Slack: Essentially globals are not really supported yet and hence can be ignored for pure functions. I will add a sentence mentioning this

Design/heap-dependent-expressions-in-pure-functions.md
mschwerhoff: That would indeed be ideal, but seems difficult to achieve for recursive predicates. But maybe I simply know too little about shared predicates. Let's discuss this in the next meeting.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
 
 ## Translation
 
-We will unfold all necessary predicates before an access. As an optimization,
-we want to keep track of previously unfolded predicates to avoid unfolding them
-again (although it is possible to do so). This ensures that there are
-sufficient permissions for all accesses made in the function. However, this
-approach comes with some limitations. To elaborate further, consider the
-following examples:
+As there is only a single expression in a pure function, we cannot simply unfold
```
mschwerhoff: Why is "on single expression" the root problem? Would this be different if multiple expressions/statements were allowed?

If it were possible to "`unfold` provided predicates from the precondition", the same could be done via `unfolding`s. But neither is possible in the context of recursive predicates, regardless of whether we are in a (single) expression context or not.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
...
+provided predicates from the precondition but rather need to place unfolding in
+expressions in the Viper encoding to ensure accesses are valid. We introduce a
+hybrid approach between greedily unfolding at the beginning of a function and 
+unfolding on-demand upon encountering a field access. Namely in SnaKt, after
+traversing a function body we have a list of `SSAAssignment`s as well as an overarching
+return expression. In the current encoding as a 'let chain' branch results are written
```
mschwerhoff: More a comment regarding the final report: Imagine you are a reader of your thesis who is not a developer of (your contributions to) SnaKt. Can you really expect them to get much out of this sentence?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
...
+to a dedicated variable and later another intermediate variable is introduced to
+resolve between the two branch results. Notice, that the branch results are defined
+regardless of whether or not the branch is used. Hence, we can unfold any predicate
```
mschwerhoff: What does it mean for a predicate to be used in a let assignment?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
...
+used in the let assignments at the beginning of the function as there will be no
+further condition that must be met in the current encoding for these predicates
```
mschwerhoff: Does that mean that predicate instances are never conditionally owned?
E.g. P(x) { ... x.f != null ==> Q(x.f) } If so, why?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
...
+to be unfolded. Limitations of this approach will be discussed later. 
+For return expressions this is different. In the current encoding, the expression
```
mschwerhoff: You initially said that there is only one expression. Doesn't that imply that these are always return expressions?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
 
 ## Translation
```
mschwerhoff: @jesyspa My comments are written with the final report in mind, so that Alex can see what kind of feedback he can expect (from me) and what I (we) expect from a technical report. Addressing these comments for the PR, whose audience is much more familiar with the technical context, may not be necessary. Please except the PR whenever it meets your requirements.

Comment regarding the final report: It is clear that this text was written by somebody who knows the implementation, but without having these additional details in the head, it is very difficult to understand, and for a reader to build a mental map of the approach.

Imagine you would give this text to another computer scientist who knows a bit of Kotlin, Viper, ASTs/CFGs. Do you think that person has a reasonable chance of applying the translation algorithm to a concrete example?

Writing technical documents is difficult, so these kinds of comments are to be expected. They still need to be addressed, though (at least for the final report).

What often helps is a somewhat formal definition of inputs, intermediate values/states, and output. E.g. you say there is a let-chain with at most one conditional in the innermost expression (by the way, where? In Kotlin's source code? In compiler's/Snakt's intermediate representation? In the Viper code you produce?). You could then specify that input to the algorithm for inferring/inserting unfolding expressions always has the shape

```
let t1 = e1.f1 in 
  let t2 = e2.f2 in 
    ...
      let tn = en.fn in 
        return b ? e' : e''
```
where
* each `e_i` is a local variable (either a function parameter `x` or a previously let-bound variable `t_j`, but in particular not a nested field access)
* `b` is a boolean expression over local variables
* `e'` and `e''` are expressions over local variables

Explications like this would give readers a much better chance at understanding what your translation algorithm is working with.

The above pseudo-formalisation might also help with elaborating how you track the predicate instances to unfold. E.g. — but I am not sure if that is what you plan to do — you could say that the algorithm traverses the let-chain from top/outermost to bottom/innermost and for each `e_i.fi`, it appends either `P(x)` if `e_i` is a function parameter, or if `e_i` is a previously let-bound variable `t_j`, then it appends  `P(p_i)` where `p_i` is the access path that corresponds to transitively unrolling the definition of `t_j` (e.g. `x.f1.f2.f3`). The latter can be assembled by also recording each let-binding (e.g. that `t1 ↦ x.f1`, `t2 ↦ t1.f2`, `t_j ↦ t2.f3`).

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -40,12 +40,26 @@ accesses" section.
...
+to a dedicated variable and later another intermediate variable is introduced to
+resolve between the two branch results. Notice, that the branch results are defined
+regardless of whether or not the branch is used. Hence, we can unfold any predicate
+used in the let assignments at the beginning of the function as there will be no
+further condition that must be met in the current encoding for these predicates
+to be unfolded. Limitations of this approach will be discussed later. 
+For return expressions this is different. In the current encoding, the expression
+of the innermost let-binding is a potentially nested Ternary expression.
+Upon encountering a field access on a variable from the let-chain
+in said return expression, we unfold any further predicates necessary. Note, that these
```
mschwerhoff: How are these "further predicates" identified? If the algorithm encounters some `t.f`, how does it know which predicate to unfold? Since `t` is a let-bound variable, the algorithm cannot simply look for a predicate instance that syntactically is some `P(t)`.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -130,12 +144,49 @@ function length(node: Ref): Int
 
 ### Limitations
 
-Note that we are acquiring more permissions than necessary. We solely infer the
-required predicates from the function's parameters. While this approach significantly
-simplifies the problem, it has some limitations. An access to a class might only occur
-under certain conditions in the function, or not at all. We still require permissions for
-the pure predicate in those cases, which could potentially cause verification
-to fail even though the function is valid.
+Note that the granularity of the provided predicate introduces some limitations
```
mschwerhoff: Which "provided predicate" do you refer to? ~~And what is its granularity? To which fields and sub-object predicates it contains permissions? The permission values (e.g. wildcard vs. full permissions)?~~ (the next sentence suggests that "granularity" means "to which fields")

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -130,12 +144,49 @@ function length(node: Ref): Int
...
+in verification capibilities. The predicate requires permissions to access all
+fields of a class. However, a function might only need access to one field of 
+said class. If in a verification context we can only infer permissions to access
```
mschwerhoff: Where is this inference happening? At call site? At implementation site (function body)? When the provided predicate is defined?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -130,12 +144,49 @@ function length(node: Ref): Int
...
+the one required field, on a conceptual level the verification should succeed.
```
mschwerhoff: Verification of ... the function call itself, i.e. establishing its precondition? Or when checking properties of the function call, i.e. about its return value? Or when the function body is verified (and its postcondition)?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -130,12 +144,49 @@ function length(node: Ref): Int
...
+    val defaultOrValue = if (node.next == null) {
+        -1
+    } else {
+        node.next.value
+    }
+    return defaultOrValue
+}
+```
+
+This will resolve to:
```
mschwerhoff: "translate"

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -130,12 +144,49 @@ function length(node: Ref): Int
...
+    unfolding purePredicate(node.next) in
+    let anon1 == (-1) in
+    let anon2 == (node.next.value) in
+    let anon3 == ((node.next == null) ? anon1 : anon2) in
+        anon3
+}
+```
+
+Clearly, this won't verify due to insufficient permissions to the purePredicate
+of node.next. Hence, this is a limitation of this approach. However note, that
```
mschwerhoff: Was the definition of `purePredicate(node)` previously shown? If not, it won't be clear why there are insufficient permissions.

Design/heap-dependent-expressions-in-pure-functions.md
jesyspa: Thank you Malte; I agree, and I think this is also useful feedback for the design as a software engineering artifact.  A structured presentation both makes it easier to understand the idea, and makes it easier to see that it is complete.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
+# Heap-dependent expressions in pure functions
+
+For a pure function, heap-dependent expressions do not violate purity per se.
+In general, as long as we only read from the heap, the function body remains
+pure. To add support for such function bodies we will rely on the previously
+introduced shared predicate. This predicate grants wildcard permissions to
+all immutable fields of a class allowing us to access those. We will require
+all necessary shared predicates in the precondition of a pure function and
+strategically use 'unfolding in' to ensure heap-accesses in the function are
+valid.
```
jesyspa: Remark: It isn't clear what this text achieves.  You essentially say that you will do the right things everywhere, but this is too vague to communicate anything about the solution.  Rather, you seem to be implicitly communicating the problem: you say "we will ... strategically use unfolding in", telling the reader that figuring out how to unfold the predicates is a question worth considering.  However, this is something that would be better communicated explicitly.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+
+## The used access predicate
+
+For now we will limit the implementation to make use of the already provided
```
jesyspa: You haven't given the reader any understanding of the problem, or the possible solutions, so saying you will "limit the implementation" does not make sense.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+shared predicate. While only granting permissions to immutable fields, this
+approach is chosen for three reasons:
+1. Convert from shared to another predicate: With the shared predicate already
```
jesyspa: "Convert from shared to another predicate" is not a reason.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+all immutable fields of a class allowing us to access those. We will require
+all necessary shared predicates in the precondition of a pure function and
+strategically use 'unfolding in' to ensure heap-accesses in the function are
+valid.
+
+## The used access predicate
+
+For now we will limit the implementation to make use of the already provided
+shared predicate. While only granting permissions to immutable fields, this
+approach is chosen for three reasons:
```
jesyspa: If you want to explain the benefit of approach A over approach B, it helps to first outline these approaches.  Alternatively, you can state your approach and then motivate it.  This style, where you explain why some one part looks like it might work but doesn't just isn't very clear.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+In this exmaple we defined another predicate (named the purePredicate) and a lemma method
+converting the shared predicate into the other predicate. We then tried to verify the
+assertion that holds in the precondition. However, this will not verify. Keeping the shared
+predicate will prevent such loss of information
+
+3. Extra permissions: Building another predicate that also supports access to
+mutable fields will cause us to inhale permissions to said fields. For these
+inhale statements there is no argument for their validity. Hence, the whole
```
jesyspa: That is not true.  If we are writing to a field, we believe that we have permissions to it. This is a reasonable assumption to pass on to Viper.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+
+The current encoding looks like the following: 
+
+let t1 := e1 in
+let t2 := e2 in
+...
+let tn := en in
+    (some condition) ? r1 : ((some_other_condition) ? r2 : r3)
+
+In this encoding ei is either a value (or an expression resolving to one),
```
jesyspa: Doesn't every expression resolve (actually: evaluate) to a value?  What do you mean by this?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+a local, a field access or a Ternary expression.
+Note however, that if ei is a Ternary expressions by definition of the encoding it
+may only have local variables (namely previously introduced let-bound variables)
+in its then or else branch. Hence, the function body
+can safely start with 'unfolding in' expressions unfolding all necessary predicates
+to grant access to all fields required in the definitions of any let-bound variable.
+The expression of the innermost let-binding is different. It is a potentially nested
+Ternary expression. ri can be any expression including field accesses on let-bound
```
jesyspa: Can the ei not use fields bound in earlier ej?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+conditions this unfold may happen.
+
+In SnaKt, we construct the previously described encoding by tracking two things:
+- A list of pairs of variable names and their definition (we call this an 
+`SSAAssignment`), where each pair will be translated into a let-bound variable.
+- A list of pairs of encountered return expressions and the conditions under which
+they are encountered. These pairs will result in the innermost Ternary expression
+described above, where the Ternary expression will resolve to the first collected
+return expression if its condition is met, to the second if the condition for the first
+is not met but the one for the second is met and so on.
```
jesyspa: This is not very clear.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+- For each `SSAAssignment` we will associate what predicates must be unfolded for
+this let-bound variable to be defined. Namely, while translating definitions of
+let-bound variables we will use the already existing 'hierarchyUnfoldPath' to
+determine the class information of any shared predicate that we have to unfold
+for this `SSAAssignment` to be valid
+- After collecting all information of the function body we will insert the
+necessary 'unfolding in' expressions directly at the field accesses in return
+expressions. As we collected the class information for all `SSAAssignemnts` we
+can infer any further predicates that need to be unfolded if a field of a
+let-bound variable is accessed.
```
jesyspa: This sounds like you are saying "we will do things right and not do things wrong, but these two bullet points together cover most of what is supposed to be in this design.  For example, how do you handle the case when the expression to be evaluated features a field read from a local variable?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+In the example below, we must also unfold the predicate of the reference to
+`next` to ensure the access made to the next node is valid. In general, nested
+field accesses can be understood as a path. When encountering a nested field
+access, we need to unfold all predicates for the types on the path that have
+not been previously unfolded. A natural consideration is to simply unfold as
+much as possible at the beginning of the function. This, however, would not
+work for the example below, as unfolding the predicates implicitly assumes that
+the path does not contain a null reference. Below, this is only true in the
+`else` branch of the ternary operator (or the Elvis operator in the Kotlin
+version); hence, an on-demand solution is required.
```
jesyspa: It is not clear from this how you decide where to put the unfolding-in.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+        node.next == null ? 
+            node.val : 
+            unfolding sharedPredicate(node.next) in node.next.val
+}
+```
+
+### Function calls to self
+
+Even though we are unfolding the predicate required by the called function,
+this translation still works. This is because, in a verification context,
```
jesyspa: This sentence only makes sense once you read the rest of the section.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,356 @@
...
+fun isLastIndex(node: Node, index: Int): Bool {
+    return length(node) == index + 1 
+}
+```
+
+```Viper
+function isLastIndex(node: Ref, index: Int): Bool 
+    requires acc(sharedPredicate(node), wildcard)
+{
+    unfolding sharedPredicate(node) in
```
jesyspa: I don't follow - why do you want to unfold here?

jesyspa (changes requested):

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
+# Heap-dependent expressions in pure functions
+
+In this document, we will discuss the design and implementation of heap-dependent
+ expressions in pure functions. We will take a look at the Viper encoding prior to 
+ this change, how to extend it to support heap-dependent expressions, and finally, 
+ what limitations arise from the designed encoding.
```
jesyspa: nit: unnecessary leading spaces

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Viper encoding prior and problem definition
+
+The current encoding translates the body of a pure function into a chain of 
+let-bindings. On the right hand side of a let-bound variable, there is, disregarding
+the Ternary guard introduced in https://github.com/JetBrains/SnaKt/pull/62, a value, 
+a local, a function call, or a  Ternary expression that only contains local variables 
+(cf. below). To extend  this encoding to support heap-dependent expressions, the
```
jesyspa, line 15: It is not clear why you are telling the reader this

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+## Viper encoding prior and problem definition
+
+The current encoding translates the body of a pure function into a chain of 
+let-bindings. On the right hand side of a let-bound variable, there is, disregarding
+the Ternary guard introduced in https://github.com/JetBrains/SnaKt/pull/62, a value, 
+a local, a function call, or a  Ternary expression that only contains local variables 
+(cf. below). To extend  this encoding to support heap-dependent expressions, the 
+following must be kept in mind:
+
+- Predicates: In SnaKt, we guard accesses to the heap using predicates. In this design,
```
jesyspa, line 18: This is rather far from true: we do not guard accesses to the heap using predicates, we maintain heap access permissions in predicates.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+let-bindings. On the right hand side of a let-bound variable, there is, disregarding
+the Ternary guard introduced in https://github.com/JetBrains/SnaKt/pull/62, a value, 
+a local, a function call, or a  Ternary expression that only contains local variables 
+(cf. below). To extend  this encoding to support heap-dependent expressions, the 
+following must be kept in mind:
+
+- Predicates: In SnaKt, we guard accesses to the heap using predicates. In this design, 
+we will rely on the shared predicate providing access to immutable fields of a class.
+While this poses some limitations, we will explain the reasoning behind this in a 
+later section.
```
jesyspa, line 21: The intended audience of this sentence is not clear, and I don't see what it accomplishes.  It essentially says "Do not worry about the ways in which this goes wrong", which isn't reassuring if you have some idea of what may go wrong, and is alarming if you don't.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+- Unfolding: For heap accesses to be valid, the corresponding predicate containing
```
jesyspa, line 22: It isn't clear how this is different from the previous point.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+While this poses some limitations, we will explain the reasoning behind this in a 
+later section.
+- Unfolding: For heap accesses to be valid, the corresponding predicate containing 
+permissions to the accessed location must be unfolded. In a pure function body, 
+we cannot use an unfold statement as we must construct a single expression (and 
+hence cannot use a statement). Therefore, we will rely on the Viper 'unfolding... in'
+expression to temporarily unfold a predicate in an expression.
+- Field access embedding: On the embedding level, field accesses are modelled 
+as FieldAccess embeddings. The receiver of these embeddings may either be a 
+variable or, in the case of a nested field access, another FieldAccess embedding.
```
jesyspa, line 29: This is an implementation detail, it is not clear why it is relevant to the design.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+In this document, we will discuss the design and implementation of heap-dependent
+ expressions in pure functions. We will take a look at the Viper encoding prior to 
+ this change, how to extend it to support heap-dependent expressions, and finally, 
+ what limitations arise from the designed encoding.
+
+## Viper encoding prior and problem definition
+
+The current encoding translates the body of a pure function into a chain of 
+let-bindings. On the right hand side of a let-bound variable, there is, disregarding
+the Ternary guard introduced in https://github.com/JetBrains/SnaKt/pull/62, a value,
```
jesyspa, line 13: "disregarding ..." is not very helpful: why should it be disregarded?  Why are you drawing attention to it if it should be?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+permissions to the accessed location must be unfolded. In a pure function body, 
+we cannot use an unfold statement as we must construct a single expression (and 
+hence cannot use a statement). Therefore, we will rely on the Viper 'unfolding... in'
+expression to temporarily unfold a predicate in an expression.
+- Field access embedding: On the embedding level, field accesses are modelled 
+as FieldAccess embeddings. The receiver of these embeddings may either be a 
+variable or, in the case of a nested field access, another FieldAccess embedding.
+
+Consider the encoding prior to this design, where ei is an expression and ri is 
+either a value or a local:
```
jesyspa, line 32: Like Malte said during our sync, presenting transformations as coming directly from surface-level Kotlin to some partly-specified intermediate syntax that you then adjust is quite confusing.  It would be much clearer to explain that you have SSAed code and need to insert appropriate scoped unfolds.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+```viper
+let t1 := e1 in
+let t2 := e2 in
+...
+let tn := en in
+    (some condition) ? r1 : ((some_other_condition) ? r2 : r3)
+```
+We want to extend this encoding to support heap accesses. Throughout this design,
+we will assume that heap accesses made in a function are pure. That is, we are
```
jesyspa, line 41: Scoping the problem - e.g. that your code is assumed to be pure - is reasonable, but it is a modification of the problem.  It is hard to understand the text if you are constantly changing the problem, so it helps to scope things up front.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+## Translation
+
+For this section, assume that the shared predicate is present in the function's context.
+We will see later how to achieve this.
+
+The core idea behind translation is to create a let-bound variable for each field access 
+and associate the necessary predicate accesses to make the field access valid with said 
+let-bound variable. Notice that the receiver of a field access is always a local variable, 
+as the receiver is either a local variable directly, or the result of translating the 
+receiver is a local variable. Whenever a new let-bound variable is introduced, we will
```
jesyspa: I don't see how I could notice this, you said yourself that FieldAccesses can have nested receivers.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+
+For this section, assume that the shared predicate is present in the function's context.
+We will see later how to achieve this.
+
+The core idea behind translation is to create a let-bound variable for each field access 
+and associate the necessary predicate accesses to make the field access valid with said 
+let-bound variable. Notice that the receiver of a field access is always a local variable, 
+as the receiver is either a local variable directly, or the result of translating the 
+receiver is a local variable. Whenever a new let-bound variable is introduced, we will 
+associate any necessary predicate accesses with it. Namely, when the following expressions
```
jesyspa: "is introduced" - by whom? You're describing this code as the result of a process you are going to hook into, but this is unclear for anyone who isn't familiar with the process.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+For this section, assume that the shared predicate is present in the function's context.
+We will see later how to achieve this.
+
+The core idea behind translation is to create a let-bound variable for each field access 
+and associate the necessary predicate accesses to make the field access valid with said 
+let-bound variable. Notice that the receiver of a field access is always a local variable, 
+as the receiver is either a local variable directly, or the result of translating the 
+receiver is a local variable. Whenever a new let-bound variable is introduced, we will 
+associate any necessary predicate accesses with it. Namely, when the following expressions 
+are let-bound to a new variable, we will do the following:
```
jesyspa: This is very hard to read compared to the corresponding set of equations.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+- Local variable: If the let-bound expression is a local variable, we will add all the 
+predicate accesses from said variable to the variable of the let-binding being introduced.
+- Field access: If the let-bound expression is a field access, we will combine the 
+predicate accesses on record for the receiver (a local variable) with the predicate 
+accesses necessary for the field access.
```
jesyspa: Is it obvious what these are, and are they always available?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+- Ternary expression: Recall that by construction, a Ternary expression will only resolve 
+between two locals. Hence, if a Ternary expression is encountered, we will associate 
+the union of the predicate accesses on record for both locals.
+- Function call: Note that a function call, by construction, will only have locals as 
+parameters. We will record the union of the predicate accesses on record for every 
+parameter as the predicate accesses associated with the let-bound variable.
```
jesyspa: Why?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+
+When let-binding an expression, we will calculate the predicate accesses 
+based on the rules above. If the let-bound expression is a field access 
+or a function call, we will wrap the expression in a sequence of unfoldings 
+in expressions, unfolding all predicate accesses calculated. The following subsections
```
jesyspa: I suspect this is the answer to "why" above: you use this list to track both what is necessary to evaluate the expression, and what is necessary to use its fields later.  It would be good to make this (and the motivation for it) explicit.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+associate any necessary predicate accesses with it. Namely, when the following expressions 
+are let-bound to a new variable, we will do the following:
+- Local variable: If the let-bound expression is a local variable, we will add all the 
+predicate accesses from said variable to the variable of the let-binding being introduced.
+- Field access: If the let-bound expression is a field access, we will combine the 
+predicate accesses on record for the receiver (a local variable) with the predicate 
+accesses necessary for the field access.
+- Ternary expression: Recall that by construction, a Ternary expression will only resolve 
+between two locals. Hence, if a Ternary expression is encountered, we will associate 
+the union of the predicate accesses on record for both locals.
```
jesyspa: Consider the following example:
```kotlin
class Box(val content: Box?)
fun f(b1: Box?, b2: Box?) : Box? {
  val v = b1?.content ?: b2?.content ?: null
  return v?.content
}
```

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+    let anon0 == (unfolding acc(sharedPredicate(node), wildcard) in node.next) in
+    let nextNode == (unfolding acc(sharedPredicate(node), wildcard) in (id(anon0))) in
+    let anon1 == (0) in
+    let anon2 == (unfolding acc(sharedPredicate(node), wildcard) in unfolding acc(sharedPredicate(nextNode), wildcard) in nextNode.value) in
+    let anon3 == (nextNode == null ? anon1 : anon2) in
+        anon3
+}
+```
+
+## The used predicate
```
jesyspa: This reads rather strangely to me, consider "Choice of predicate"

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+{
+    convertSharedToPure(head)
+    assert unfolding acc(pure_predicate(head), wildcard) in head.val == 42
+}
+```
+
+In this example, we defined another predicate (named the purePredicate) and a 
+method converting the shared predicate into the other predicate. We then tried 
+to verify the assertion that holds in the precondition. However, this will not 
+verify. Keeping the shared predicate will prevent such loss of information:
```
jesyspa, line 208: nit: colon?

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+Instead of using the shared predicate, which only grants permissions to immutable 
+fields of a class, one could have created another predicate, say the pure predicate, 
+granting wildcard permissions to all fields of a class. We decided against this for
+the following reasons:
+
+1. Complexity of converting the shared predicate to another: With the shared predicate 
+already being present, to obtain another predicate, one would typically convert the 
+shared predicate into another predicate using a lemma function or method. As the shared 
+predicate is class specific, it is non trivial to define one (or multiple) Lemma 
+functions handling this conversion.
```
jesyspa, line 160: Just for the record, this isn't really true.  The amount of effort is comparable to the addition of the havoc methods.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+
+predicate purePredicate(curr: Ref) {
+    acc(curr.val, wildcard) &&
+    acc(curr.next, wildcard) &&
+    (curr.next != null ==> acc(purePredicate(curr.next)))
+}
+
+method convertSharedToPure(curr: Ref)
+    requires acc(sharedPredicate(curr), wildcard)
+    ensures acc(purePredicate(curr), wildcard)
```
jesyspa, line 185: Since there's no condition relating the old and the new state here, it will not verify.  I think this doesn't quite cover why you couldn't add that, though.

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Constructing the precondition of a function
+
+Note that all reference types for which we need to require the shared predicate 
+can be found in the function's parameters. Furthermore, because the predicate 
+also contains the predicates for any references within the parameters, no 
+extra predicates need to be acquired to ensure that potentially nested heap 
+accesses in the function body work.
+
+## Using a function
```
jesyspa, line 218: This section is kind of strange - it is long and doesn't seem to say much beyond "this already works"

Design/heap-dependent-expressions-in-pure-functions.md
```diff
@@ -0,0 +1,360 @@
...
+{
+    let nextNode == (getNext(node)) in
+    let anon0 == (nextNode.value) in
+    let anon1 == (0) in
+    let anon2 == (nextNode == null ? anon1 : anon0) in
+        anon2
+}
+```
+
+In the let-binding of `anon0` the verification context is not holding sufficient
```
jesyspa, line 396: Interesting point, so to what degree does calling functions within functions work?

jesyspa (approved):
I have left quite a few comments.  These are all non-blocking; I think it is better to merge some version of this and then keep going, rather than trying to get it perfect now.

That said, I think the issues regarding ternaries are not solved, and an implementation of this proposal would not generally work.

PR #6, opened 2026-03-06
Permission Management

The-Ray-Man:
This PR updates the design on how the permission management (fold/unfold) should be performed.

### DIFF
--- a/Design/folding-unfolding-while.md
+++ b/Design/folding-unfolding-while.md
@@ -1,0 +1,98 @@
+## Folding-Unfolding-While
+
+
+### Assumptions
+In the first iteration we make the following simplifying assumptions:
+- We only read from datastructures.
+- Per iteration, we go at most one level deeper.
+- We only consider completely unique datastructures.
+
+
+### Observations
+- We can imagine the datastructures as trees.
+- There can be a **constant** number of "runners". Runners are variables that are 
+assigned to some nodes of the tree and might go deeper in every loop iteration. 
+E.g. for a linked list the ``current`` would be a runner.
+- If in the current loop iteration, a runner is moved deeper into the tree, we will call 
+this runner active. To refer to the initial runner of a loop iteration, we will call them base runner.
+- If the object is borrowed, we need to carry the permissions through the loop.
+Meaning after the loop we need to fold everything back until we reach the root.
+- We will need to use the magic wand feature from viper. 
+- The invariant of the loop must be the magic wand giving us the root of the tree.
+As well as the predicates from the runners.
+- The magic wand should have the form of the unique predicate.
+  - (foreach field: field != null ==> unique(field)) --* unique(root)
+- At the end of a loop iteration, we need to update the magic wand. This update 
+usually involves the following steps:
+  -  Create a new magic wand, equivalent with the one in the invariant.
+  -  As the body, 
+      - For every active runner, add the necessary fold statements to go back to the base runner.
+      - apply the old magic wand (for all active runners use the corresponding base runner). 
+
+
+  
+### Examples
+
+
+
+#### Linked List Traversal
+Some pre/postconditions as well as invariants were removed for better readability.
+````viper
+method contains(list: Ref, p$v: Ref)
+  returns (ret$0: Ref)
+  requires acc(UniqueLinkedList(list), write)
+  ensures acc(UniqueLinkedList(list), write)
+{
+    var current: Ref
+    var anon$0: Ref
+    var current_old: Ref
+
+    unfold acc(UniqueLinkedList(list), write)
+    current := list.bf$head
+
+    // start of preping the loop.
+    // runners need to be traded for the root.
+    package (current != nullValue() ==> acc(UniqueNode(current), write)) --* acc(UniqueLinkedList(list), write) {
+        fold acc(UniqueLinkedList(list), write) 
+    }
+    
+    label lbl$continue$0
+    invariant (current != nullValue() ==> acc(UniqueNode(current), write)) --* acc(UniqueLinkedList(list), write)
+    invariant (current != nullValue() ==> acc(UniqueNode(current), write))
+
+    anon$0 := sp$notBool(boolToRef(current == nullValue()))
+    if (boolFromRef(anon$0)) {
+        var anon$1: Ref
+        unfold acc(UniqueNode(current), write)
+        anon$1 := current.bf$value
+        if (intFromRef(anon$1) == intFromRef(p$v)) {
+            ret$0 := boolToRef(true)
+            // fold here, because current is no longer moved.
+            // since we do not have any active runners, no need to update the magic wand.
+            fold acc(UniqueNode(current), write)
+            goto lbl$ret$0
+        }
+        // a runner update happens. we need to save the old runner.
+        current_old := current
+        // current is already unfolded.
+        current := current.bf$next
+        
+        // we have seen a runner update. So we need to update the magic wand.
+        package (current != nullValue() ==> acc(UniqueNode(current), write)) --* acc(UniqueLinkedList(list), write) {
+            // in this loop iteration we unfolded current_old. So this needs to be reversed.
+            fold acc(UniqueNode(current_old), write)
+            // apply the old magic want (updated current with current_old)
+            apply (current_old != nullValue() ==> acc(UniqueNode(current_old), write)) --* acc(UniqueLinkedList(list), write)
+        }
+        goto lbl$continue$0
+    }
+    label lbl$break$0
+
+    ret$0 := boolToRef(false)
+    
+    goto lbl$ret$0
+    label lbl$ret$0
+    // after the loop we can get the root back again.
+    apply (current != nullValue() ==> acc(UniqueNode(current), write)) --* acc(UniqueLinkedList(list), write)
+}
+````
--- a/Design/folding-unfolding.md
+++ b/Design/folding-unfolding.md
@@ -1,56 +1,115 @@
 # Predicates folding/unfolding
 
 ## Expected Behaviour
+Deciding what has to be unfolded can be quite involved. The following classes should include most cases.
+
+```kotlin
+class A(
+    @Unique @Borrowed var mu : B,
+    @Unique val iu : B,
+    @Unique @Borrowed var ms : B,
+    val is : B
+)
+
+class B(
+    @Unique @Borrowed var mu : C,
+    @Unique val iu : C,
+    @Unique @Borrowed var ms : C,
+    val is : C
+)
+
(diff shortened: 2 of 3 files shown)


Design/outdated/folding-unfolding.md
```diff
@@ -0,0 +1,95 @@
...
+    - If the `PredCtx` contains a unique-predicate that after several unfoldings can provide write access to `p`, these
+      unfoldings are performed.
+    - otherwise access is inhaled.
+- Assignments `... = p`. We need to access `p` with at least read permissions, there are 3 alternatives:
+    - The `PredCtx` contains `acc(p, write)` or `acc(p, wildcard)` -> nothing to do.
+    - If the `PredCtx` contains a predicate that after several unfoldings can provide access to `p`, these unfoldings
+      are performed.
+    - otherwise access is inhaled. **NOTE**: it is **not** possible to inhale it with `wildcard` permissions since we
+      want to exhale it right after the read (and there is no way to exhale `wildcard` access). The best thing to do is
+      inhaling it with a small fractional permission.
```
jesyspa, line 36: Isn't this the case in which we havoc?

jesyspa (changes requested):
I think a core downside here is that you want to generate `PredCtx` from the ExpEmbedding, but this is a significant duplication of work that is going to already be done in the uniqueness checker.  Is there a way that the context could be constructed from the uniqueness information more directly?

Design/outdated/folding-unfolding.md
The-Ray-Man: Yes that is true
(this is the old folding strategy written by Francesco. I moved it into the outdated folder instead of deleting it, because I am not sure If I captured all the valuable insights form it in the new file)

jesyspa: Ohh, sorry, I totally missed that.

Design/folding-unfolding-while.md
```diff
@@ -0,0 +1,98 @@
...
+E.g. for a linked list the ``current`` would be a runner.
+- If in the current loop iteration, a runner is moved deeper into the tree, we will call 
+this runner active. To refer to the initial runner of a loop iteration, we will call them base runner.
+- If the object is borrowed, we need to carry the permissions through the loop.
+Meaning after the loop we need to fold everything back until we reach the root.
+- We will need to use the magic wand feature from viper. 
+- The invariant of the loop must be the magic wand giving us the root of the tree.
+As well as the predicates from the runners.
+- The magic wand should have the form of the unique predicate.
+  - (foreach field: field != null ==> unique(field)) --* unique(root)
```
jesyspa, line 24: One thing I've been thinking: would having a separate shared/unique predicate for nullableX that represents "X or null" be helpful?  It sounds like some things could be more uniform this way.

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+
+* The color of the nodes is the result of the uniqueness type system.  
+* Circles represent objects, while rectangles represent the fields of those objects.  
+* The Uniqueness Type text inside a box is the declared uniqueness type of that field.  
+* An arrow from ``object -action-> field`` indicates the `action` required to access that `field` on the `object`.  
+* To access a deeper field, follow the arrows and execute the actions in the order of traversal.  
+* To prevent the diagram from becoming too large, we use dotted arrows to indicate that we have reached a state equivalent to one already seen.
+```mermaid
+graph LR
+    A_u((Class A <br/> Unique)) -- unique --> B_um_ui["um<br/>Mutable / Unique <br/> ui<br/>Immutable / Unique"]
```
jesyspa, line 36: nit: the blue is very hard to read, maybe it can be a bit lighter?

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+class Bb_si_ui shared
+class Bb_sm_um shared
+
+```
+
+
+
+#### Mealy State Machine
+
+The process of determining necessary actions can be expressed as a Mealy state machine. The input is the accessed path. The action unfold means that we need to unfold the path read so far (excluding the field responsible for the transmission). The first transmission is performed according to the uniqueness of the receiver.
```
jesyspa, line 95: Oh, that's cool!

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+    A -.-> Shared
+    B -.-> Shared
+```
+
+
+(Due to an issue with the rendering engine, extra nodes were inserted in the diagram above.)
+
+## Takeaways
+
+* If the receiver is shared, we will never unfold a unique predicate. This is convenient because shared predicates do not need to be folded back.
```
jesyspa: I don't see the connection between these two statements, could you elaborate?

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+* To prevent the diagram from becoming too large, we use dotted arrows to indicate that we have reached a state equivalent to one already seen.
+```mermaid
+graph LR
+    A_u((Class A <br/> Unique)) -- unique --> B_um_ui["um<br/>Mutable / Unique <br/> ui<br/>Immutable / Unique"]
+
+    A_s((Class A <br/> Shared)) -- shared --> Bb_si_ui["si<br/>Immutable / Shared <br/> ui<br/>Immutable / Unique"]
+    A_s -- havoc --> Bb_sm_um["sm<br/>Mutable / Shared <br/> um<br/>Mutable / Unique"]
+
+
+    B_um_ui -.-> info23("Equivalent with A Unique")
```
jesyspa, line 42: When you say "equivalent", what is equivalent to what?

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+* si: Shared and Immutable
+
+### Visualization
+
+The diagram should be read as follows:
+
+* The color of the nodes is the result of the uniqueness type system.  
+* Circles represent objects, while rectangles represent the fields of those objects.  
+* The Uniqueness Type text inside a box is the declared uniqueness type of that field.  
+* An arrow from ``object -action-> field`` indicates the `action` required to access that `field` on the `object`.
```
jesyspa: action: unfold that predicate?

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+
+```Kotlin
+// Snippet 1  
+var b = (root.left == null)
+
+// Snippet 2  
+var tmp = root.left
+```
+
+In the first snippet, we want to unfold unique(root) once and then fold it back. In the second, we only want to unfold unique(root) because root.left is moved to tmp.
```
jesyspa, line 158: I agree this is a difficult question.  The main solution I've thought about was to have cases like this take a local reference.  Do you want to do something more elaborate via the cfg?

Design/folding-unfolding.md
```diff
@@ -1,95 +1,203 @@
...
+3. Extract all prefixes containing only Unfold-Unique actions.  
+4. Filter for uniqueness and order them by length.  
+5. Insert the unfold statements.
+
+## Unique Folding Strategy
+
+In the folding strategy, we only need to consider unique predicates. Shared predicates are unfolded with wildcard permissions, meaning we retain access to the predicate. This works closely with the Unique Unfolding strategy:
+
+1. Identify all used paths.  
+2. Extract all prefixes that are unique and not partially moved.
```
jesyspa, line 189: I think I see, but I would really prefer one of two approaches (depending on what's reasonable):
1. Reads know whether they are moving or not
2. Reads have a lifetime for which they are moving

In the first case, you know whether to fold or not after a read.  In the second, you know to consider re-folding at the end of the lifetime.

jesyspa (approved):
One thought: I get the impression that maybe this repository is not that convenient for working with designs :/ I feel like the iteration process is quite slow, because context is lost.  May be worth thinking of using google docs instead, even though those do get lost much more easily.

Approving this so that you can fix small issues and bigger ones can be treated as follow-ups.  For the while stuff, I need to think more.

Design/folding-unfolding-while.md
The-Ray-Man: I think that could simplify some things, but also complicate some others. How are smart casts or in more general smart casts handled. I think we would need to exchange the `uniqueNullableX` predicate for a `uniqueX` predicate. Not sure how many new edge cases would be introduced with more predicates available.

Design/folding-unfolding.md
The-Ray-Man: Could you elaborate a bit more, how local reference would solve this problem?
In this example I try to demonstrate, why given the information form the CFG, we can not easily perform the folding of unique predicates on the `FieldAccess` level. I do not plan to do complicate things with the cfg. I just looked at how the information from the CFG can be translated into the FIR elements and later in the ``ExpEmbeddings``. If one does this in a straight forward way, then the paths become moved at the location of assignment and not the read location (which I think is the correct way to think about it).
Generally I would be a bit hesitant to perform more involved CFG to FIR translation. For example: One solution to this problem would be to perform a look a head in the CFG to figure out whether the field access results in a moved path or not. With this approach we would be able to know if after the field access we need to fold back or not. However I belive if we using strategies like this, then we introduce many edge-cases where this could go wrong.

Design/folding-unfolding.md
The-Ray-Man: Generally, I think option 2 would be more reasonable. I think that would require a extension to the data flow analysis (@paganma).

I would suggest the following:
- I start to implement the proposed system. And test it rigorously. 
- In the meantime marco could look into how to figure our the lifetimes of the reads. 
- Given the new information we update the folding logic. Which I think should not be very different than this.
  - If a lifetime ends, we check if the owner is now no longer moved. If it is not moved -> fold back
  - On the statement level, we need to see if the "written to path" is no longer mvoed. If it is not moved -> fold back

Design/folding-unfolding.md
jesyspa: Sorry, I didn't phrase that very well.  What I meant is that if `==` has type `(local, local) -> Bool` then we don't need to worry about the first case.  The second case is more complex, but can also be reasoned about on the type level (by introducing lifetimes and then unifying them).  I agree it should be a unified solution and not just a number of special cases.

Design/folding-unfolding.md
jesyspa: That sounds good to me.  I think detecting the end of a move can be a bit difficult, we may need to do something special there, but as a starting point we could look at the defining scope of a variable.

Design/folding-unfolding-while.md
jesyspa: That's a good point, I'll bring this up with Athena

PR #7, opened 2026-03-09
Uniqueness Data-flow Analysis

paganma:
This PR introduces a semi-formal specification of the dataflow analysis used to evaluate uniqueness types across a Kotlin control-flow graph.

### DIFF
--- a/Uniqueness/system-proposals/dataflow-analysis.typ
+++ b/Uniqueness/system-proposals/dataflow-analysis.typ
@@ -1,0 +1,235 @@
+#title("Uniqueness Data-Flow Analysis")
+
+This document defines a forward, intra-procedural data-flow analysis for uniqueness over Kotlin FIR control-flow graphs.  At each control-flow node, the analysis tracks an abstract environment mapping paths to uniqueness facts. A fact records whether the path is `unique` or `shared`, whether it is `local` to the current function or may escape the current function, and whether the path has been `moved`.
+
+= Syntax Definition
+<syntax>
+
+We assume we are operating on the standard Kotlin FIR CFG nodes defined in:
+
+```kotlin
+package org.jetbrains.kotlin.fir.resolve.dfa.cfg
+```
+
+The syntax described below is an abstract syntax for the CFG shapes relevant to this analysis.
+
+== Paths
+<paths>
+
+Paths are syntactic units representing variable locations and field locations. The syntax can be described as a list of symbols:
+
+#let Path = $italic("Path")$
+#let Symbol = $italic("Symbol")$
+#let path = $p$
+$
+    Symbol & ::= x | y | z | ... \
+    Path path & ::= Symbol | Symbol . Path
+$
+
+A single-component path represents a local variable, while a multi-component path represents a field access.
+
+#let subpaths = $italic("subpaths")$
+
+For a path $path$, the set $subpaths(path)$ contains $path$ itself together with every descendant field path rooted at $path$.
+
+== Uniqueness Types
+<uniqueness-type-lattice>
+
+#let BorrowLevel = $italic("Borrow-Level")$
+#let UniqueLevel = $italic("Unique-Level")$
+#let blevel = $b$
+#let ulevel = $u$
+
+Uniqueness types describe the aliasing constraints known for a path at a given CFG point. A usable path has two components:
+/ The aliasing status of the path: described by $UniqueLevel$. A path is #emph[unique] if no other accessible alias of that path may exist. A path is #emph[shared] if aliases may exist.
+/ The locality of the path: described by $BorrowLevel$. A path is #emph[local] if it may not be aliased outside the current dynamic scope. A path is #emph[global] if it may escape the dynamic scope.
+
+Additionally, the lattice contains a distinguished element #emph[moved]. This denotes a path that is temporarily or permanently unavailable because its unique value has been moved away or is inaccessible during a borrow. The set of possible abstract values is:
+
+#let Type = $italic("Type")$
+#let type = $t$
+#let global = $mono("G")$
+#let local = $mono("L")$
+#let moved = $mono("M")$
+#let shared = $mono("S")$
+#let unique = $mono("U")$
+$
+    Type type & ::= (UniqueLevel, BorrowLevel) | moved \
+    BorrowLevel blevel & ::= global | local \
+    UniqueLevel ulevel & ::= unique | shared
+$
+
+The partial order $subset.eq$ is defined by the following relations:
+$
+    (unique, global) & subset.eq (unique, local) \
+    (shared, global) & subset.eq (shared, local) \
+    (unique, global) & subset.eq (shared, global) \
+    (unique, local) & subset.eq (shared, local) \
+    (shared, local) & subset.eq moved
+$
+
+The transitive closure of $subset.eq$, together with the induced greatest lower bound (meet) $inter$ and least upper bound (join) $union$, provides the lattice used by the analysis.
+
+== Control-Flow Nodes
+
+We assume we are operating on a generic Kotlin control-flow graph representing the body of a function declaration.
+
+#let fun = $mono("fun")$
+#let node = $n$
+#let predecessors = $italic("predecessors")$
+#let successors = $italic("successors")$
+
+Every node $node$ of the graph has a set $successors(node)$ of immediate successor nodes and a set $predecessors(node)$ of immediate predecessor nodes. We designate nodes using the expressions and statements to which they correspond. The syntax relevant to this analysis is the following:
+
+#let Assignment = $italic("Assignment")$
+#let Call = $italic("Call")$
+#let Declaration = $italic("Declaration")$
+#let Expression = $italic("Expression")$
+#let Node = $italic("Node")$
+#let Parameter = $italic("Parameter")$
+#let Statement = $italic("Statement")$
+#let expression = $e$
+$
+    Node node & ::= Declaration | Parameter | Expression | Statement | Call | Assignment | * \
+    Declaration & ::= fun Path( Parameter ... ) { Statement ... } | * \
+    Parameter & ::= Path: Type \
+    Expression expression & ::= Path | Call | * \
+    Statement & ::= Assignment | Call | * \
+    Call & ::= Path(Expression ...) \
+    Assignment & ::= Path = Expression
+$
+
+Other CFG constructs, including branches and loops, are handled through the predecessor and successor relation of the underlying CFG together with the join operator defined below.
+
+== Evaluation Context
+
+#let hole = $cal("E")$
+Each source-level statement or expression may correspond to multiple CFG nodes. To describe the exact evaluation point, we use an evaluation hole denoted by $hole$. A hole pattern is written as $hole[Expression]$ and matches the subexpression currently being evaluated while preserving its surrounding context. For example, consider the assignment:
+$
+    v = m(f(x, y, z))
+$
+
+The pattern $v = hole[expression]$ matches the following CFG nodes in dominance order:
+$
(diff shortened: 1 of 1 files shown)


Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+Envout(s) &= traverse(s) quad \
+Envin(s) &= join(pred(s))
+$
+where $pred(s)$ is the set of control-flow predecessors of $s$; the function $traverse(s)$ defines the effect of each kind of statement; the function $join({p_1,...,p_n})$ combines the outgoing environments of all predecessors into a single environment that will be the incoming environment of $s$.
+
+== Traverse Function
+<sec:traverse>
+The $traverse(s)$ function describes how the execution of a single statement modifies the output environment $Envout$ in relation to the input environment $Envin$.
+
+#let hole = $cal(E)$
```
quartztz: This hole sort of comes out of nowhere in the document, it's unclear what the notation below really means. It can probably be inferred by staring at the identities long enough, but it could be worth to explain the meaning of cal(E) and the traverse function a bit more?

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+#let Path = $italic("Path")$
+#let Type = $italic("Type")$
+$ 
+Env : Path -> Type
+$
+We write $Env[x]$ for the type of path $x$ in $Env$, and $Env[x |-> t]$ for the environment that agrees with $Env$ except at $x$, where it maps to $t$.
+
+= Environment Initialization
+<environment-initialization>
+The initial environment $Env$ flowing through each intermediate statement will be the empty environment $nothing$. The environment flowing through the first statement $Env_0$ should reflect the method's parameters' specifications. For example, for the following method declaration:
```
quartztz: Somewhat unclear what "method" this refers to: if the analysis is done per method frame that should be made clearer in the document. This sentence is very short anyways, so a formal definition of what the statement we're considering is could be useful.

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+```kotlin
+fun m(unique Ref x, unique local Ref y)
+```
+The flow object for the starting statement will be
+$ 
+Env_0 = { x |-> (unique, global), y |-> (unique, local) } 
+$
+
```
quartztz: The example is much appreciated!

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+traverse(hole[x : (unique, global)]) & = Envin[x |-> moved] \
+traverse(hole[x : (unique, local)]) & = Envin[x |-> moved] \
+traverse(hole[x : (shared, global)]) & = Envin \
+traverse(hole[x : (shared, local)]) & = Envin[x |-> moved] \
+traverse(x = e) & = Envin[x |-> type(e)] \
+traverse(f(nothing)) & = Envin \
+traverse(f(e dot E)) & = restore(e) union traverse(E)
+$
+
+// TODO: Define hole, type, restore
```
quartztz: I had not seen this comment, so I guess it makes sense It wasn't defined earlier. Still i think it would be good to define the notions soon.

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+
+== Join of Predecessor Environments
```
quartztz, line 210: This section is very clear! Understandable and well formulated.

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,111 @@
...
+Additionally the lattice contains a top element #emph[moved] that denotes a unique path which has been moved, and hence cannot be used.
+
+Formally, we denote the set of types as:
+#let unique = $mono("U")$
+#let shared = $mono("S")$
+#let global = $mono("G")$
+#let local = $mono("L")$
+#let moved = $mono("M")$
+$ 
+& "Type" = {\
```
quartztz: While it is very funny, the terminology Consumed and Borrowed does not appear anywhere else in the document (the first section prefers global and local, as the variable names and initials would imply). Is this a relic of old terminology or is it deliberate?

quartztz (approved):
A few questions about some formulations or terminologies. Other than that, the document reads well, the procedure is understandable enough and the main mechanisms are well presented!

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,187 @@
...
+We use the function $subpaths(path)$ to retrieve the subcomponents of a path.
+
+== Uniqueness Types
+<uniqueness-type-lattice>
+#let UniqueLevel = $italic("Unique-Level")$
+#let ulevel
+#let BorrowLevel = $italic("Borrow-Level")$
+#let blevel
+Uniqueness types describe the constraints on the shareability of a path at any given point in the execution of the control-flow graph. Informally, the uniqueness of a path is characterized by:
+/ The aliasing status of the path : described by $UniqueLevel$. If a path is #emph[unique] no other alias of that path can exist. If a path is #emph[shared] it may have been aliased by another path.
```
jesyspa: There is a bit of nuance here: there may be aliases to this path up the stack (that have been borrowed), but there may not be aliases that are accessible to other functions (e.g. via globals).

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,187 @@
...
+
+== Uniqueness Types
+<uniqueness-type-lattice>
+#let UniqueLevel = $italic("Unique-Level")$
+#let ulevel
+#let BorrowLevel = $italic("Borrow-Level")$
+#let blevel
+Uniqueness types describe the constraints on the shareability of a path at any given point in the execution of the control-flow graph. Informally, the uniqueness of a path is characterized by:
+/ The aliasing status of the path : described by $UniqueLevel$. If a path is #emph[unique] no other alias of that path can exist. If a path is #emph[shared] it may have been aliased by another path. 
+/ The ability to alias the path : described by $BorrowLevel$. If a path is #emph[local] it cannot be aliased further outside of the current function. If a path is #emph[global] it may be assigned elsewhere.
```
jesyspa: This is a bit of a hard one to phrase.  Concretely: local annotates the variable with an (implicit) lifetime, and the subtyping is contravariant in lifetime inclusion.

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,187 @@
...
+
+== Transfer Function
+<sec:transfer>
+The $transfer(node)$ function describes how the execution of a single statement modifies the output environment $envout$ in relation to the input environment $envin$ with the following rules:
+#let type = $italic("type")$
+#let reroot = $italic("reroot")$
+#let var = $italic("var")$
+$
+transfer(node) = envin \
+transfer(path_1  = hole[path_2 : (unique, blevel)]) & = envin[path_2 |-> moved] union {path_1 . path_3 |-> envin[path_2 . path_3] | path_3 in subpaths(path_2)} \
```
jesyspa: Why is this true?  Maybe I don't understand how the evaluation context works, but this reads to me as that path 2 can be transformed before being assigned to path 1?

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,187 @@
...
+
+Support for other operations, including branch and loop statements, is implied by their immediate dominance relationship.
+
+== Evaluation Context
+#let hole = $cal(E)$
+Note that for each statement and expression in the program there may be multiple control-flow nodes. To enable a finer-grained description of the current execution step we rely on the notion of evaluation hole denoted as $hole$. A hole pattern takes as input a pattern matching the expression that is currently being evaluated using the form $hole[Expression]$. This pattern allow us to match control-flow nodes while also referencing the context. For example, consider the assignment:
+$
+v = m(f(x, y, z))
+$
+For this statement, the pattern $v = hole[expression]$ matches the following control-flow nodes preserving the dominance order $->$:
```
jesyspa: I don't quite follow - are contexts a syntactic convenience, or a part of the CFG structure, or what?  (Maybe it would be easier to discuss over a call)

Uniqueness/system-proposals/dataflow-analysis.typ
```diff
@@ -0,0 +1,187 @@
...
+== Transfer Function
+<sec:transfer>
+The $transfer(node)$ function describes how the execution of a single statement modifies the output environment $envout$ in relation to the input environment $envin$ with the following rules:
+#let type = $italic("type")$
+#let reroot = $italic("reroot")$
+#let var = $italic("var")$
+$
+transfer(node) = envin \
+transfer(path_1  = hole[path_2 : (unique, blevel)]) & = envin[path_2 |-> moved] union {path_1 . path_3 |-> envin[path_2 . path_3] | path_3 in subpaths(path_2)} \
+transfer(path_1 = expression) & = envin[path_2 |-> default(path_1)] \
```
jesyspa: Is path_2 a typo here?

jesyspa (changes requested):
Sorry for the delay - finally got around to reading this.  I'm not quite sure I understand the formalism you have in mind

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: Specified that this is an intra-procedural analysis at the start of the document.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: This is indeed a relic. Right now we just switched to "locality" terminology.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: Noted. Now I state that a local variable may not be aliased *outside* of the current function.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: `path_2` is moved because it sits at the tail of the right-hand of the assignment. This node actually refers to the access to `path_2`, which indeed happens before the assignment.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: They are a syntactic convenience. In practice the way I would obtain the context is by traversing the successor nodes of the path access nodes until the corresponding assignment is found.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: I have now specified this with the aid of a `copy` function which assists in rerooting a path within another.

paganma: In the third rule path_2 looks like a typo, yes.

Uniqueness/system-proposals/dataflow-analysis.typ
paganma: My understanding is that, since we lack support for rust-style lifetimes, the locality of a variable limits its ability to flow into a variable that was defined in another lexical scope or a closure.

paganma: I'm not sure whether this generalization works.

jesyspa (approved):
Sorry for the delay on this, it looks good to me.  I think it's clear on the current approach, and while I'm sure questions will come up during formalization, we shouldn't block the merge of this on that.

PR #11, opened 2026-04-14
Immutable Fields to Functions

The-Ray-Man:
Discussion tracking issue #12


Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
+# Kotlin Immutable Fields in Viper
+The core goal is to map Kotlin fields to Viper based on whether they are stable (pure) or heap-dependent. By using the static type at the access point, we decide whether to use a Viper function, a method, or a direct field access.
```
jesyspa: > they are stable (pure) or heap-dependent

I think this mixes a few things:
- stable: a property described by the Kotlin spec, depends on notions like packages
- pure: a semantic property of the function
- heap-dependent: a property of the Viper representation

I think it would be good to be clearer here.

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Stability through Viper Functions
+For fields that are guaranteed to be immutable and final (closed), we use Viper functions.
+- Advantage: Functions are pure and do not require heap permissions (acc) to evaluate.
+- Construction: The relationship between the constructor argument and the field is established in the constructor's postcondition: `ensures A_field(res) == field_val`.
+- Subtype Specificity: For covariant overrides, the function uses implications to refine the return type: `ensures subtype(this, SubClass) ==> subtype(result, SpecificType)`
+
+## Static Type Dispatch Logic
+Instead of generating complex ternary case distinctions at every call site, the verifier chooses the access method based on the static type of the field.
```
jesyspa: I don't quite follow, what "ternary case distinctions" do you mean?  (In general, it's good to not assume the previous state is clear; designs are read long after the previous implementation is forgotten.)

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
...
+|---------------------------------------	|-----------------	|--------------------------------------------------------------------------------------------------------------------------	|
+| Closed val                            	| Function Call   	| The value is stable and guaranteed not to be overridden by a mutable implementation.                                     	|
+| Open val                              	| Method Call     	| Since it's open, a subclass might implement it with a custom getter or a var, requiring a method to abstract the access.	|
+| Closed var                            	| Viper Field     	| Direct heap access is used. This requires the verifier to manage permissions.                                 	|
+| Open var                              	| Method Call     	| To support dynamic dispatch and potential setters in subclasses, we wrap the access in a method.                         	|
+| Getter exists | Method Call | If there is a custom getter, then we always must use a method.
```
jesyspa: There's an inconsistency here: you treat open cases separately (though they're the same), but here you make a case that overlaps the others.

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
...
+
+Since the types can be covariantely overwritten, we need to make sure that the necessary type information is present. However in this example this is not really relevant.
+```viper
+
+method sides_open(this: Ref) returns (ret: Ref)
+    ensures subtype(ret, Int) // default, because of open
+
+function sides_closed(this: Ref) returns (ret: Ref)
+    ensures subtype(this, Triangle) ==> subtype(ret, Int)
+    ensures subtype(this, Square) ==> subtype(ret, Int)
```
jesyspa: Interesting idea.  I think there's two things worth considering here:
1. Can `sides_closed` be defined with the `ensures` defined separately?  If so, how?
2. Can we somehow ensure "if `sides_open` is called with `this subtype T`, then the result is equal to `sides_closed(this)`"?

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
...
+            verify(r1==r2) // this will fail, because we are missing information from the first read.
+            }
+    }
+}
+```
+
+
+At the moment there is a missing connection between multiple reads if the type information in between changed. Respectively, when we use the `field_open` read, we learn nothing about the the potential values of different types. For example in the example above the equality `r1 == r2` should hold. However at the moment this would not verify. 
+
+### Suggestion 1: Add Subtype Distinction in Open Read Method
```
jesyspa: The issue here is that you end up with a lot of places where information has to be aggregated.  Maybe that's reasonable, but I'm not sure.

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,258 @@
...
+We could also add postconditions for final mutable fields, that say, if it is of this type and the receiver has write permissions(which means that it is unique), then the result will be the same as in the viper field.
+
+This would then look like this:
+```viper
+field f_Mutable: Ref
+
+method field_open(this: Ref) returns (ret: Ref)
+    ensures subtype(ret, Int)
+    ensures subtype(this, Mutable) ==> subtype(ret, Int)
+    ensures subtype(this, Mutable) && [perm(this.f_Mutable) == write ==> ret == this.f_Mutable, true]
```
jesyspa: Does this work?  I would expect you to need a magic wand.

jesyspa (approved):
Thanks, I think this makes a lot of sense.

My main overarching feedback is that instead of putting multiple things in the postconditions of one method, I'd sooner try to split it out.  But maybe this isn't actually necessary.

Design/Draft/immutable-fields.md
The-Ray-Man: Yeah, I am sorry. I had an additional section, which I divided to remove in the end.

Design/Draft/immutable-fields.md
The-Ray-Man: Which cases are overlapping?
In the new version I updated the table. I think now it should be correct.

Design/Draft/immutable-fields.md
The-Ray-Man: 1. Yes we can. We can put it into an axiom. Not sure though if that is much nicer. But I only see these two options. There is a section on the new proposal on this.
2. Yes this is possible as well. There are some paragraphs in section 2 of the new proposal.

Design/Draft/immutable-fields.md
The-Ray-Man: The short examples I tried worked. But not sure how well it is generalized. [Here](https://viper.ethz.ch/viperonline/tool/silver?key=2026042014596f37a92c9fd214f19f9085bb) would be some examples.

The-Ray-Man:
I updated the proposal and tried to implement your feedback and divide the idea into three parts. 

> My main overarching feedback is that instead of putting multiple things in the postconditions of one method, I'd sooner try to split it out. But maybe this isn't actually necessary.

About this: if we go with the second part of the proposal, then we would have one method per field. This method would have one postcondition for each object that overwrites the field. These would be of the form:
`ensures isType(this, X) ==> ret == field_closed(this)`
The type information for the read value can be stored in an axiom.

Design/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+First, we want to see to which viper construct different kind of field accesses should be mapped. 
+
+| Field Characteristic (at Static Type) 	| Viper Construct 	| Reasoning                                                                                                                	|
+|---------------------------------------	|-----------------	|--------------------------------------------------------------------------------------------------------------------------	|
+| Open val/var                              	| Method Call     	| Since it's open, a subclass might implement it with a custom getter or a var, requiring a method to abstract the access.	|
+| Closed val*                            	| Function Call   	| The value guaranteed not to be overridden.                                     	|
+| Closed var*                           	| Viper Field     	| Direct heap access is used. This requires the verifier to manage permissions.                               	|
+
+*For these two cases we also need to analyze getters. If there are getters defined by the user, then this getters have to be called instead of the call to the function/access to the viper field.
```
jesyspa, line 28: For vars, also setters?

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+For the immutable field `val immut` we create a viper function:
+```viper
+function field_immut_closed(this: Ref) : (ret: Ref) 
+    ensures isType(ret, Int)
+
+method test(a: Ref){
+    var l_immut = field_immut_closed(a)
+}
+```
+We can ensure the exact type, because there is no possiblity of overwritting the field with a subtype. Viper function are always pure, which means that given the same receiver it will always return the same return value.
```
jesyspa: I don't think this is true in general.  Yes, the field will always be of that type, but the actual value may be of a subtype.  Of course, Int has no subtypes, but consider a field of type Any.

We don't generally track exact types, it doesn't add much in terms of verification possibilities and is more brittle.

Design/Draft/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+We have `Y <: X` and `Z <: X` but neihter `Y <: Z` nor `Z <: Y`
+
+
+For every field, we need to check if it appears as a open, closed or both. In this example all fields appear as open and closed. Meaning we need both the function and method (for immutable fields) or method and field (for mutable fields).
+
+For the field `property` we need the following viper constructs:
+```viper
+function field_property_closed(this: Ref) : (ret: Ref)
+    ensures isType(this, Triangle) ==> subtype(ret, Y)
+    ensures isType(this, Square) ==> subtype(ret, Z)
```
jesyspa: Is it worth bundling them like this, instead of using separate functions for the separate subtypes?  I don't see much benefit, and it seems like it complicates things if one subtype has a val and another a var.

Splitting it also allows
```
function Triangle_field_property_closed(this)
  requires isType(this, Triangle)
  ensures subtype(...)
```

Design/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+
+```viper
+method test(x: Ref) {
+    var prop_shape = field_property_open(x)
+    if (*) {
+        // x is Triangle
+        var prop_triangle := field_property_closed(x)
+        
+        // unfold
+        x.field_color := RGB(255,0,255,1)
```
jesyspa, line 193: Well, this is assuming some uniqueness, right?  As is, we throw this write away.

Design/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+
+## 3. Represent Every Field Read as Method Call
+The last stage of the design propsal is to also the reading of mutable + closed fields as method calls. Then every reading field access would be the same method call. 
+
+For this the postcondition of the `open` method must be updated:
+
+```viper
+method field_color_open(this: Ref) returns (ret: Ref)
+    ensures subtype(ret, RGB)
+    ensures isType(this, Triangle) ==> [perm(this.f_Mutable) == write ==> ret == this.field_color, true]
```
jesyspa, line 300: As I said in the meeting, I am worried that this is all very conditional.

Design/immutable-fields.md
```diff
@@ -0,0 +1,305 @@
...
+    forall this: Ref :: { field_property_closed(this) }
+        isType(this, Triangle) ==> subtye(field_property_closed(this), Y)
+        isType(this, Square) ==> subtye(field_property_closed(this), Z)
+}
+
+function field_property_closed(this: Ref) : (ret: Ref)
+
+method field_property_open(this: Ref) : (ret : Ref)
+    ensures subtype(ret, X)
+    ensures isType(this, Triangle) ==> ret == field_property_closed(this)
```
jesyspa, line 288: It's a pity that the connection to the method can't be moved to an axiom, but maybe it's not that bad.

jesyspa (approved):
Other than the point on exact types, this LGTM

Design/immutable-fields.md
The-Ray-Man: That is a good question. I guess it depends on how, the setter is implemented. If there is a setter, then this setter should be a method which takes the permissions to the field. If the setter does not have any pre/postconditions, then after the setter method is called the value of the field will be unknown. Hence when reading from it (i.e. reading the viper field), we should get a unknown value. 
But afaik in the current state, setters are not implemented like this.

Design/Draft/immutable-fields.md
The-Ray-Man: No, I think there are not really any benifits from it. I will split them

Design/immutable-fields.md
The-Ray-Man: yes, I will add a comment in the beginning of the file that this is a general assumption we take.

PR #14, opened 2026-06-29
CFG Analysis for Fold/Unfolds

The-Ray-Man:


Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+
+class B(
+    var a1: @Unique A
+    var a2: @Unique A
+)
+
+```
+
+
+### Unused Reads
```
jesyspa, line 33: Hmm, I'm a bit confused.  For both it should be okay to fold b, since x is an alias of b.a (but you lose the accesses x has if you fold b).

Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+```
+The expected viper would look like this:
+
+```viper
+anon := con_A(5)
+unfold(Unique(anon))
+x := anon.field
+```
+
+But if we perform the naive statement level approach, we would extract the paths, see that for the field access we need to unfold the receiver and then add the unfold statement with the receiver. Which would result in about following viper code:
```
jesyspa, line 64: This is actually not a problem unique to this: `x ?: y` also has similar issues.  But I agree it isn't nice.

Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+# Problem
+- We want to add the fold/unfolds on the field access expression. But during the translation from Fir to ExpEmbedding, we do not know if it needs to be folded or unfolded. The main reason is, that to fold back, we need to know if after the current statement is finished the path is partially moved or not.
+
+## When to fold and unfold
+I think the best way to connect the uniqueness information to the predicate state is the following:
+- If variable is unique, then we have access to the predicate.
+- If a variable is partially moved, then the predicate is unfolded.
+
+## When to decide?
+I think the best place to decide the folds+unfolds is in the translation from the `Fir` to `ExpEmbedding`. Since, the uniqueness analysis runs on the `Fir` it generates facts based on the `Fir` embeddings. So to use this information, we need to query it with `Fir` embeddings. If we do it in the `ExpEmbedding` or even later, we would need some way to extract the initial `Fir` embedding, which would be error prone and cumbersome.
```
jesyspa, line 15: I think this groups things that may be different.  That is: we can consume the FIR-based uniqueness information into something that describes the permissions we have/want, without committing to specific folds/unfolds.

Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+```viper
+unfold(unique(b))
+consume(b.a1, b.a2)
+```
+Doing this on a statement level would be finde, because the path extractor will realize that there is a common prefix and only unfold it once. However if we move the fold/unfold decision to the field access, then it becomes unclear how to stop the double unfolds.
+
+During translation of these two field accesses the uniqueness information will be the same for both. Hence they will perform the same unfold.
+
+
+### Nested Function Calls
```
jesyspa, line 102: This particular example would break actually (there's a way to alias unique pointers if we allow it).  But we'll see when we get further with uniqueness.

Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+    val x = (if (cond) {
+        val y = c1.b
+        y.a1
+    } else {
+        c1.b.a1
+    }).field
+}
+```
+This is just a nightmare...
+This example breaks the current uniqueness checker. It reports a uniqueness violation even thought the program is correct.
```
jesyspa, line 142: Hmm, that's strange.  I should look into the implementation at some point. :/

Design/folding-unfolding-cfg-analysis.md
```diff
@@ -0,0 +1,361 @@
...
+
+This example also illustrates why we can not surface the folds to the statement level. The field access `y.a1` would be surfaced to the outermost assignment. But in that context, the `y` which should be unfolded does not even exist.
```
jesyspa, line 144: I agree.  I think we should derive the folds during linearization, when we have more info.

jesyspa (commented):
Some thoughts, going to need more time with the later part.

